PR #19, opened 2020-11-13
Video dataloader

RaivoKoot:
Added one new file at `kale.loaddata.video_dataset.py` with very comprehensive ReadTheDocs docstrings.  
Added example in `examples.video_dataset_loading` that demonstrates using this for loading and preprocessing/augmentation of a tiny dummy dataset. A proper description of all of this is the README of the example project. Please take a look at that one for documentation.  
  
Note: alltogether added images of tiny dummy dataset are very small (400KB)


examples/video_dataset_loading/main.py
```diff
@@ -0,0 +1,109 @@
...
+if __name__ == '__main__':
+    """
+    This demo uses the dummy dataset inside of the folder "demo_dataset".
+    It is structured just like a real dataset would need to be structured.
+    """
+    videos_root = os.path.join(os.getcwd(), 'demo_dataset')
+    annotation_file = os.path.join(videos_root, 'annotations.txt')
+
+    """ DEMO 1 WITHOUT IMAGE TRANSFORMS """
+    dataset = VideoFrameDataset(
```
haipinglu, line 17: Will import kale.loaddata.video.frameset as VideoFrameDataset be better? I am not sure. We can discuss tomorrow.

examples/video_dataset_loading/main.py
```diff
@@ -0,0 +1,109 @@
...
+        plt.show()
+        plt.pause(1)
+
+
+    """ DEMO 2 WITH TRANSFORMS """
+    # As of torchvision 0.8.0, torchvision transforms support batches of images
+    # of size (BATCH x CHANNELS x HEIGHT x WIDTH) and apply deterministic or random
+    # transformations on the batch identically on all images of the batch. Any torchvision
+    # transform for image augmentation can thus also be used  for video augmentation.
+    preprocess = transforms.Compose([
```
haipinglu, line 44: I am not sure. We have tried to put transforms under `kale.prepdata` earlier.

examples/video_dataset_loading/main.py
```diff
@@ -0,0 +1,109 @@
...
+    # tensor of shape (NUM_SEGMENTS*FRAMES_PER_SEGMENT) x CHANNELS x HEIGHT x WIDTH
+    frame_tensor = sample[0]
+    print('Video Tensor Size:', frame_tensor.size())
+    # integer label
+    label = sample[1]
+
+
+    def denormalize(video_tensor):
+        """
+        Undoes mean/standard deviation normalization, zero to one scaling,
```
haipinglu, line 72: For what?

haipinglu (changes requested):
Personally, I think "*data*" in several names is not necessary, e.g. 
kale.loaddata.video_dataset.py vs kale.loaddata.video.py  
examples.video_dataset_loading vs examples.video_loading 
See more examples in pytorch, e.g. torchvision.datasets.CIFAR100 not torchvision.datasets.CIFAR100_dataset
But torch also has repetition like yours, e.g. torch.utils.data.TensorDataset, torch.utils.data.ChainDataset 
I am not sure so we can discuss. We should learn from [`torch.utils.data` API](https://pytorch.org/docs/stable/data.html) & [`torchvision.dataset` API](https://pytorch.org/docs/stable/torchvision/datasets.html) without losing our (hopefully) good characteristics.

I agree that the demo data is small, 405KB in size (492 KB on disk of my laptop). How about our repository size? It is currently only 238KB so the data has 1.7x the 238KB. Also this PR is said to have "*42*" files to review (which sounds intimidating) but actually 35 are those images not for review. Will others working on graphs interested in having this data? If you are removing it, possible to remove the commits so that it does not go into the history? We can discuss tomorrow.

Actually, this is what Peizhen did recently in his GripNet PR. See the [GripNet ReadMe](https://github.com/pykale/pykale/tree/master/examples/drug_gripnet). His demo data has a size of 151KB only but it is still put on Google Drive. Actually pytorch has API to auto-download Google Drive data as well. Simple rules are simple to follow. Otherwise, we need to define clearly how small is small to allow in `examples`. Some may think 10M or 100M is small. I revised the pykale guideline to reflect this.

From the recent lesson, I hope to keep the pykale repository as data-free as possible. We should separate data and code if possible. I hope data can be hosted externally, e.g. on (your or pykale group's) GitHub or Google Drive, and will be downloaded to a fixed directory `examples/data` as needed, just as CIFAR10. This also allows us to use larger data too.

haipinglu:
Or maybe there is no need for a separate example purely on video loading. Just merge with Xianyuan's action_dann? Not sure, to discuss.

RaivoKoot:
> I agree that the demo data is small, 405KB in size (492 KB on disk of my laptop). How about our repository size? It is currently only 238KB so the data has 1.7x the 238KB. Also this PR is said to have "_42_" files to review (which sounds intimidating) but actually 35 are those images not for review. Will others working on graphs interested in having this data? If you are removing it, possible to remove the commits so that it does not go into the history? We can discuss tomorrow.

True, I didn't realize the repo was only 238KB. Thought it was 6MB. In that case I'll update the PR so that my demo downloads and extracts the demo dataset from elsewhere instead of containing it.

haipinglu:
To my understanding, 238KB is the size of the master (the future release
version for `pip install`) and 6MB is the whole repository size including
all commits history (for `clone`).

examples/video_dataset_loading/main.py
RaivoKoot: When you use a dataset class with transforms and then try to access a sample, you receive the sample after the transforms are applied. So, in the beginning each frame of a sample consists of an integer tensor with values betwen 0 and 255, but when you access that sample, each frame is scaled to range [0,1] and then  normalized to mean zero and standard deviation one before you receive it. After that you cannot plot the frames anymore for visualization, because plotters only know how to show images either in the range [0,1] or [0,255].  
  
So, typically you need to denormalize correctly back to [0, 255] or [0,1] before being able to do any visualization like I do below those lines.

examples/video_dataset_loading/README.md
```diff
@@ -0,0 +1,139 @@
+# Efficient Video Dataset Loading, Preprocessing, and Augmentation
+Author: [Raivo Koot](https://github.com/RaivoKoot)  
+  
+If you are completely unfamiliar with loading datasets in PyTorch using `torch.utils.data.Dataset` and `torch.utils.data.DataLoader`, I recommend
+getting familiar with these first through [this](https://pytorch.org/tutorials/beginner/data_loading_tutorial.html) or 
+[this](https://github.com/utkuozbulak/pytorch-custom-dataset-examples).
```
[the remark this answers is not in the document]
RaivoKoot: > To my understanding, 238KB is the size of the master (the future release version for `pip install`) and 6MB is the whole repository size including all commits history (for `clone`).

Yeah I realized that now too. I will for sure remove the dataset from the repository itself.

RaivoKoot:
I think you're right about some of the redundant naming. I will revise the PR naming to something that seems better and let you know when I am ready for a new inspection of that.

kale/loaddata/video_dataset.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def __len__(self):
+        return len(self.video_list)
+
+def imglist_totensor(img_list):
+    """
+    Converts each PIL image in a list to
+    a torch Tensor and stacks them into
+    a single tensor. Can be used as first transform
+    for ``kale.loaddata.video_dataset.VideoFrameDataset``.
```
RaivoKoot, line 237: I initially felt like it would be better to have this specific transform in here but have changed my mind and agree that putting it into prepdata is better. Will make the change

RaivoKoot:
I've update things now to address your good comments.
1. Removed dummy dataset from repo and made it download-extract on-demand when running main.py. 
- [ ]  Am also on email with David to make sure the commits including the image files get removed from my branch/PR commit history so that they don't increase the repo size. Will not merge until that is done.

2. Moved the one video transform to a new module kale.prepdata.video_transform, as ImglistToTensor(), analogous to torchvisions ToTensor() for single images.
3. Renamed module file to kale.loaddata.videos and the example to examples.video_loading. Changed readTheDocs title to "VIDEOS - Data Loading". I have left VideoFrameDataset naming as is, because generic torch.utils.Dataset implementations seem to all use this naming convention. Only dataset specific ones like CIFAR10 or so don't use "dataset" in the name I think.
  
I think in this state, it is ready to get merged (after correcting git history). Based on Xianyuan and my conversation, we will then, in a new PR/branch, update VideoFrameDataset to incorporate the functionality that his dataloaders offer right now but not mine. And because of that will not include his dataloaders, if possible. In a comment below, I will post the email conversation between him and me.

RaivoKoot:
-----------------------------
From Xianyuan, 17:08:
For now, I have some different requirements as below.
1. We have four subsets, which are GTEA, KITCHEN, ADL and EPIC-Kitchen. 
The former three are small (I think they are small now. Maybe in the experiment, they are large enough. Who knows, Hahaha😂) Thus, I plan to use all frames from the video. In this setting, your current Video Frame Sampling Method is not suitable for me. We may make some changes on it for our dataset. 

However, for EPIC-Kitchen, it is too large, and I plan to select some frames from each action sample or each video. In this setting, your data loader is a good choice. For now, I follow a paper to select 16-frame continuous frames from each action video randomly.

2. In my opinion, I hope we can use the same series of data loaders to maintain the consistency of PyKale, even if there are different requirements. Thus, if you are willing to do it, we can extend your code to suit our dataset. We can use similar codes to get the data loader, but there are also some challenges I can think of now:
a. My image file is img_{:010d}.jpg.
b. Numbering may not start at 1.
c. Some need the sampling part, while others don't. They need to load 16-frame continuous frames.
d. Annotation files are different. Mine is [Video_name, start_frame, end_frame, action, verb, verb_label], while yous is [Video_name/start_frame, length_of_frames, label]

If you are interested in making it for our dataset, I am glad to discuss and do it with you.

-------------------------------
From Raivo, 18:50:
Good job on outlining/listing the differences/requirements. Yeah I agree with you on (2). It might be too much to all fit in a single dataset class, but I'm very willing to try and integrate the extra features we need into the dataset and hopefully address them all. In the best case scenario, my class can fit all requirements and we don't need anything else, in the close-to-best case, we have my class implement most requirements and have a separate class for some edge cases or so maybe.

I can luckily address a few of your challenges already now:
a: The image file names are not a problem. When instantiating my class, you can pass in the template, such as "img_{:010d}.jpg. So that is not a problem
b: I think this might be an easy fix for me, but will take some extra conversation.
c. This can be achieved with my class. When you set the NUMBER_OF_SEGMENTS=1, the whole video is treated as a single segment. And if you then set FRAMES_PER_SEGMENT=16, 16 continuous frames will be sampled from the video with a random start index. I should add some documentation into the README for this and add a demonstration of this into the main.py file. Good for pointing this out though. I only realize now that my class can actually do this haha.
d1. Good point. I forgot to think about situations where the labels are not just 1 integer, like in EPIC KITCHENS. I'll think about the best way to update this to accommodate arbitrary label formats. 
d2. Also, I'll think about ways to maybe change my expectation of the annotation file format further. If instead of asking for the number of frames, I ask for start and end frame, that would fix issue b and also allow multiple dataset samples to come from the same video that lies in the same folder with all frames (like in EPIC KITCHENS). And if I do it like this, I can still use it with my current usage, by setting start frame to 1 and end frame to NUM_FRAMES. 
z: This is a new point of mine. Right now, I don't support anything else but rgb frames. I don't know how flow frames are usually dealt with so when we speak I can ask you a few things about how they are used and see whether I can fit this into my class too.

Nice. I'll try to first address haiping's comments on my PR and then try to add the clear features from above that I think should go in the class for sure. When I'm done with that, I'll message you and we can set up a meeting to think about the more difficult requirements like flow and maybe any others that come to mind!  
  
-------------------------------
Xianyuan very happily agrees with this

RaivoKoot:
> Or maybe there is no need for a separate example purely on video loading. Just merge with Xianyuan's action_dann? Not sure, to discuss.

Small answer to this comment: I think newcomers and even just people who are new to this dataloader implemenation will desperately need a simple/independent example like this. Theres often too much distracting and confusing code if something like this is instead demonstrated inside a whole system implementation. I think this standalone example will therefore be very needed and helpful.

examples/video_loading/main.py
```diff
@@ -0,0 +1,133 @@
...
+    frame_tensor = sample[0]
+    print('Video Tensor Size:', frame_tensor.size())
+    # integer label
+    label = sample[1]
+
+
+    def denormalize(video_tensor):
+        """
+        Undoes mean/standard deviation normalization, zero to one scaling,
+        and channel rearrangement for a batch of images.
```
haipinglu, line 97: From your reply, then shall we add something like "for proper visualisation" or " so that we could visualise the frame/video properly"? E.g. new comers may have similar question if they are not familiar.

haipinglu (commented):
Just finished reading all your update. Good job and teamwork. I fully agree keeping the video load example separate. It will be very useful indeed. I just made one minor comments about clarifying the purpose of denormalise in docstrings and no further comment on other issues. Once the data file commits are removed. You or @XianyuanLiu can do the merge.

kale/loaddata/video_dataset.py
haipinglu: Yeah, as long as it is **general** and highly **reusable**, we should put it in core.

examples/video_dataset_loading/main.py
haipinglu: Thanks! Clarifying this in the docstrings may be better.

examples/video_dataset_loading/main.py
haipinglu: Still in main now, right?

examples/video_loading/README.md
```diff
@@ -0,0 +1,139 @@
+# Efficient Video Dataset Loading, Preprocessing, and Augmentation
+Author: [Raivo Koot](https://github.com/RaivoKoot)  
+  
+If you are completely unfamiliar with loading datasets in PyTorch using `torch.utils.data.Dataset` and `torch.utils.data.DataLoader`, I recommend
+getting familiar with these first through [PyTorch data loading tutorial](https://pytorch.org/tutorials/beginner/data_loading_tutorial.html) or 
+[PyTorch dataloading examples](https://github.com/utkuozbulak/pytorch-custom-dataset-examples).
+### Overview: This example demonstrates the use of `kale.loaddata.video_dataset.VideoFrameDataset`
```
haipinglu, line 7: now "loaddata.videos"

examples/video_loading/main.py
RaivoKoot: Yeah good point. I'll add that

examples/video_loading/README.md
RaivoKoot: of course thanks for pointing out

PR #24, opened 2020-11-19
Clean the requirements so that they are installable

drj11:
This makes it possible to install pykale with the following:

    git clone https://github.com/pykale/pykale
    cd pykale
    python -m pip install -r requirements.txt

(assuming that you have a suitable Python install already)

I've commented out the dependencies for torch-geometry because they can't be installed at the same time as `torch`, they have to be installed after `torch` is installed. But we need to address that (it's possible that switching to setup.cfg or an alternate approach would work).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,11 +4,11 @@
 # We currently recommend Python 3.6 or above.
 
 # core dependencies
-torch==1.5.0
+torch==1.7.0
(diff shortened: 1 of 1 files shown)


RaivoKoot:
updated to require torchvision 0.8 instead of 0.6 and torch 1.7 instead of 1.5

PR #25, opened 2020-11-23
Update mpca.py

sz144:
Updating mpca documentation with examples. Fixed the bug where error reported if the first eigenvalue already expalined the variance needed.


kale/embed/mpca.py
```diff
@@ -1,49 +1,76 @@
...
-        ----------
-        Attributes:
-            tPs: list of projection matrices
-        """
-        self.var_exp = variance_explained
+    """MPCA implementation compatible with sickit-learn
+
+    Args:
+        var_explained (float, optional): ration of variance to keep
+            (between 0 and 1). Defaults to 0.97.
```
haipinglu, line 28: Typo ration --> ratio, maybe using percentage will be better. See https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html

kale/embed/mpca.py
```diff
@@ -93,38 +129,57 @@ def fit(self, X):
...
+def MPCA_(X, var_explained=0.97, max_iter=1):
+    """MPCA implementation as an independent function
+
+    Args:
+        X (ndarray): training data, shape (I1, I2, ..., n_samples)
+        var_explained (float, optional): ration of variance to keep (between 0 and 1). Defaults to 0.97.
+        max_iter (int, optional): max number of iteration. Defaults to 1.
+
+    Returns:
+        list: a list of projection matrices
```
haipinglu, line 170: Shouldn't this be a the transp_proj_mats?

kale/embed/mpca.py
```diff
@@ -93,38 +129,57 @@ def fit(self, X):
...
+        """
+        
+        return multi_mode_dot(X, self.tPs, modes=[m for m in range(self.ndim - 1)], transpose=True)
+
+
+def MPCA_(X, var_explained=0.97, max_iter=1):
+    """MPCA implementation as an independent function
+
+    Args:
+        X (ndarray): training data, shape (I1, I2, ..., n_samples)
```
haipinglu, line 165: Will *X_train* be better? As in https://scikit-learn.org/stable/auto_examples/applications/plot_face_recognition.html

kale/embed/mpca.py
```diff
@@ -93,38 +129,57 @@ def fit(self, X):
...
-    """
-    Parameter:
-        X: array-like, ndarray of shape (I1, I2, ..., n_samples)
-        variance_explained: ration of variance to keep (between 0 and 1)
-        max_iter: max number of iteration
-    ----------
-    Return:
-        tPs: list of projection matrices
+        Returns:
+            ndarray: Data in original shape, shape (I1, I2, ..., n_samples)
```
haipinglu, line 155: Is there a variable name?

kale/embed/mpca.py
```diff
@@ -93,38 +129,57 @@ def fit(self, X):
...
-        ----------
-        Return:
-            Data in original shape, array-like, shape (I1, I2, ..., n_samples)
-        """
-        return multi_mode_dot(X, self.tPs, modes=[m for m in range(self.n_dim - 1)], transpose=True)
+        """Transform data in the shape of reduced dimension 
+        back to the original shape
 
+        Args:
+            X (ndarray): shape (i1, i2, ..., n_samples), where i1, i2, ...
```
haipinglu, line 150: Reduce from I1 to i1 may seem weird, particularly to multilinear algebra notations. I usually use P1, ..Pn, following conventions.

kale/embed/mpca.py
```diff
@@ -93,38 +129,57 @@ def fit(self, X):
...
-        ----------
-        Return:
-            Transformed data, array-like, shape (i1, i2, ..., n_samples)
+        """Perform dimension reduction on X
+
+        Args:
+            X (ndarray): shape (I1, I2, ..., n_samples)
+
+        Returns:
+            ndarray: Transformed data, shape (i1, i2, ..., n_samples)
```
haipinglu, line 138: Is there a variable name?
I seem to understand why. If this is what sklearn does, it is fine.

kale/embed/mpca.py
```diff
@@ -93,38 +129,57 @@ def fit(self, X):
...
-        """
-        Parameter:
-            X: array-like, shape (I1, I2, ..., n_samples)
-        ----------
-        Return:
-            Transformed data, array-like, shape (i1, i2, ..., n_samples)
+        """Perform dimension reduction on X
+
+        Args:
+            X (ndarray): shape (I1, I2, ..., n_samples)
```
haipinglu, line 135: Is there a more specific name for X such as X_high_dim or X_original or X_orig

kale/embed/mpca.py
```diff
@@ -1,49 +1,76 @@
...
-            X: array-like, ndarray of shape (I1, I2, ..., n_samples)
-        ----------
-        Return:
-            self
+    def fit(self, X, y=None):
+        """Fit the model with X.
+
+        Args
+            X (ndarray): input data, shape (I1, I2, ..., n_samples), where n_samples
+                is the number of samples, I1, I2 ... are the dimensions of corresponding
```
haipinglu, line 58: I1, I2, .... does not specify when to end.
Will I_1, ..., I_N or I1, I2, .... be better?

kale/embed/mpca.py
```diff
@@ -1,49 +1,76 @@
...
-        self.var_exp = variance_explained
+    """MPCA implementation compatible with sickit-learn
+
+    Args:
+        var_explained (float, optional): ration of variance to keep
+            (between 0 and 1). Defaults to 0.97.
+        max_iter (int, optional): max number of iteration. Defaults to 1.
+
+    Attributes:
+        tPs: list of projection matrices
```
haipinglu, line 32: Yes, I used it in the Matlab code but now it is time to make it **more readable** like those in sklearn https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html . 't' stands for transposed (see matlab code _tUs{n}=Us{n}'_. Therefore, we can call it _transposed_proj_mats_ or _trans_proj_mats_ or _transp_proj_mats_

The doc can also be improved. In matlab 
tUs: the multilinear projection, consisting of N projection matrices, one for each mode
Will it be better to mention the same that there are N of them, one for each mode?

kale/embed/mpca.py
```diff
@@ -93,38 +129,57 @@ def fit(self, X):
         return self
 
     def transform(self, X):
```
haipinglu, line 131: Do we have to use all bases? Can I input a lower dimension such as 2, 5, 10?

kale/embed/mpca.py
```diff
@@ -1,49 +1,76 @@
...
-        Parameter:
-            X: array-like, ndarray of shape (I1, I2, ..., n_samples)
-        ----------
-        Return:
-            self
+    def fit(self, X, y=None):
+        """Fit the model with X.
+
+        Args
+            X (ndarray): input data, shape (I1, I2, ..., n_samples), where n_samples
```
haipinglu, line 57: Will X_train be better?  Not sure.

kale/embed/mpca.py
```diff
@@ -1,49 +1,76 @@
...
-        """
-        self.var_exp = variance_explained
+    """MPCA implementation compatible with sickit-learn
+
+    Args:
+        var_explained (float, optional): ration of variance to keep
+            (between 0 and 1). Defaults to 0.97.
+        max_iter (int, optional): max number of iteration. Defaults to 1.
+
+    Attributes:
```
haipinglu, line 31: It seems that the sorting order (odrIdx in matlab) is not returned. If we fit to 100 but want to test performance w.r.t. 10, 20, 30, how to find the first 10/20/30 Multilinear PCs? With odrIdx, I can just take the first 10/20/30. With the current code, do I have to fit again three times?

haipinglu (changes requested):
Overall it looks good and it is good to aim for sickit-learn compatibility. 
I identified several possible ways of improvement. If anything is not clear, we can discuss when we meet to sort it out. Thanks.

kale/embed/mpca.py
sz144: Yes, we should make it clearer and reduce barriers here. I followed the scikit-learn documentation here (https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html#sklearn.decomposition.PCA.inverse_transform). But it can be confusing for some users. I will change it later.

kale/embed/mpca.py
sz144: Shall we use X_original for input (high-dimensional) data and X_transformed for output (low-dimensional) data

kale/embed/mpca.py
sz144: Yes, will change

kale/embed/mpca.py
sz144: 1. Do you mean we can have mode-wise dimension reduction here?

2. The input data shape (P_1, ..., P_N) must be consistent with the training data. I will improve the documentation later.

kale/embed/mpca.py
sz144: How about making odrIdx an attribution of MPCA object? Just like tPs in the current implementation.

kale/embed/mpca.py
sz144: Will change later.

kale/embed/mpca.py
haipinglu: Good to see you are following top examples. Then maybe that's OK. Of course, sometimes, we can do better if we have a strong justification.

kale/embed/mpca.py
haipinglu: Not necessarily, right? We can apply MPCA on transformed data as well. Maybe I am getting overcomplicated and keeping as simple as X is the best.

kale/embed/mpca.py
haipinglu: Sorry, I mean Will I_1, ..., I_N or I1, I2, ....I_N be better? (missed I_N)

kale/embed/mpca.py
haipinglu: Maybe just *proj_mats*? We can explain it is transposed in docstrings.

kale/embed/mpca.py
haipinglu: Not necessarily mode-wise. When we do MPCA, e.g. we have an **output** size of 20x20x20. When I use it, I may want a 10x10x10, or just a vector of 30. The current code cannot do that. And without ordidx in the matlab code, it is difficult to get a vector feature out. We can talk if you are still not clear. 

The input data size should not be P_1, .., P_N, it should be I_1, ..., I_N, right?

kale/embed/mpca.py
haipinglu: Yes, I think it is good. This also highlights our deep understanding of MPCA. Without ordIdx, MPCA won't be flexible. 

Ranked/ordered features are a good characteristic of PCA/MPCA, e.g. compared to ICA or Autoencoder, in which the features do not have a natural ordering. 

Of course, ordIdx is actually a univariate feature selection process (unsupervised version by default, sort by variance). But PCA has a natural ordering so it will be good for MPCA to have it naturally.

kale/embed/mpca.py
```diff
@@ -1,181 +1,308 @@
...
+    Args:
+        var_explained (float, optional): amount of variance to keep
+            (between 0 and 1). Defaults to 0.97.
+        max_iter (int, optional): max number of iteration. Defaults to 1.
+
+    Attributes:
+        proj_mats (list): a list of transposed projection matrices
+        idx_order (ndarray): the ordering index of projected (and vectorised) features in decreasing variance
+        shape_in (tuple): input tensor shapes, i.e. (P_1, P_2, ..., P_N)
+        shape_out (tuple): output tensor shapes, i.e. (i_1, i_2, ..., i_N)
```
haipinglu, line 78: I am confused here. Typically, we use I_1, ..., I_N to denote the input dimensions and P_1, ..., P_N to denote the output dimensions. Why using opposite notations here?
Also i-n should be capitalised to I_n.

kale/embed/mpca.py
```diff
@@ -1,181 +1,308 @@
...
+        self.ndim = X.ndim
+        self.shape_in = shape_[:-1]
         self.Xmean = np.mean(X, axis=-1)
         X_ = np.zeros(X.shape)
         # init
         Phi = dict()
         Us = dict()  # eigenvectors
-        Vs = dict()  # eigenvalues
+        # Vs = dict()  # eigenvalues
+        lmds = dict()  # eigenvalues
```
haipinglu, line 134: Make it more readable: lambdas

kale/embed/mpca.py
haipinglu: The question above is not implemented yet. What if I want to get a lower dimension of 19? Let's talk in our meeting if you are not clear.

haipinglu (changes requested):
Almost there. Thanks.
1. Why swap I_n and P_n conventions for input and out dimensions.
2. Capitalise dimension I_1 not i_1
3. Enable `transform( )` to get MPCA of any lower dimension using `idx_order`, e.g. 19. 

If you are not clear, we can talk when we meet.

kale/embed/mpca.py
sz144: Fixed. Please see the latest update.

kale/embed/mpca.py
sz144: fixed in the last commit

kale/embed/mpca.py
```diff
@@ -178,52 +181,65 @@ def _fit(self, X):
...
         Args:
-            X (ndarray): shape (P_1, P_2, ..., P_N, n_samples)
+            X (ndarray): Data to perform dimension reduction, shape (I_1, I_2, ..., I_N, n_samples).
+            vectorise (bool): Whether vectorise the transformed tensor. Defaults to Flase.
+            n_components (int): Number of components to keep. Applies only when vectorise=True. 
+                Defaults to None.
 
         Returns:
-            ndarray: Transformed data, shape (i_1, i_2, ..., i_N, n_samples)
+            ndarray: Transformed data, shape (P_1, P_2, ..., P_N, n_samples).
```
haipinglu, line 210: That's if `vectorise=false`.
If `vectorise=true`, the shape should be (`n_components`, `n_samples`), right?

kale/embed/mpca.py
```diff
@@ -146,23 +149,23 @@ def _fit(self, X):
         for i in range(self.ndim - 1):
             eig_vals, eig_vecs = np.linalg.eig(Phi[i])
             idx_sorted = eig_vals.argsort()[::-1]
-            Us[i] = eig_vecs[:, idx_sorted]
+            Vs[i] = eig_vecs[:, idx_sorted]
```
haipinglu, line 152: We can use a more interpretable name for `Vs`, e.g. `sorted_eig_vecs` or something else more readable.

kale/embed/mpca.py
```diff
@@ -1,181 +1,324 @@
...
 
-    return tPs
+        Returns:
+            ndarray: Data in original shape, shape (I_1, I_2, ..., I_N, n_samples)
+        """
+        _check_dim_shape(X, self.ndim, self.shape_out)
+        return multi_mode_dot(X, self.proj_mats, modes=[m for m in range(self.ndim - 1)], transpose=True)
 
 
+# The following implementation kept for a tensorly function
```
haipinglu, line 242: What's the purpose of keeping this? And it is not synchronised with recent update.

haipinglu (approved):
Thanks for the update. It looks great now. Three minor comments:
1. `Vs` can be changed to something more readable.
2. Docstring for returning dimension for `transform` should consider the flexible case, vector or tensor, full or partial projection. 
3. Unclear about the purpose of the commented `MPCA_` for `tensorly`, which is not synchronised either. Is it still useful to keep it there?

I will merge since the above issues are minor but please take a look and fix directly if appropriate.

kale/embed/mpca.py
sz144: should be (n_samples, P_1 * P_2 * ... * P_N) or (n_samples, n_components), which is consistent with scikit-learn pca, will update the docstring soon

kale/embed/mpca.py
sz144: will remove in the next update

kale/embed/mpca.py
sz144: will do

PR #27, opened 2020-12-10
improved videoloader for usage with more diverse datasets

RaivoKoot:
# Added some functionality to videoloader
I only really made a few low level changes to the videoloader code. Its not very reviewable. Most line changes are just in the README and in code comments.

### Added
Videoloader now supports multiple labels per sample like NOUN, VERB, ACTION in EPIC-KITCHENS. Also now supports specifying START and END frame per sample like in EPIC-KITCHENS. A demo of this new functionality is also included in the demo file `main.py`  
  
These are changes that we needed so that Xianyuan could use this loader for many different datasets that he uses (and just make it more flexible in general for anyone of us).


haipinglu (approved):
As you have expected, I have no comments made after going through the changes. This is a solid piece of work. Well done!

PR #32, opened 2021-01-05
Update seed.py(have reverted) & solve the previous issues 

XianyuanLiu:
### 1. Overview
I have updated `set_seed` in `kale.utils.seed` (No need) & solve the previous issues in the example `action_dann_lightn`.

### 2. Description
a. Improved `set_seed` to set the seed for generating random numbers for GPU. (No need)
b. Added some comments to `action_dann_lightn`.


XianyuanLiu:
No need to update seed.py because `torch.manual_seed(seed)` does `torch.cuda.manual_seed_all(seed)` inside. I will update comments with the next PR.

PR #28, opened 2020-12-29
Add flow branch on I3D

XianyuanLiu:
### 1. Overview
I have added optical flow branch to the example `action_dann_lightn`.

### 2. Description
Main work is on loading and converting optical flow frames to make it available for training on I3D. However, it is not perfect and remains some problems as below. 
a. Only work on I3D architecture because I haven't found the optical flow pre-trained R3D and R2+1D.
b. Only work on RGB or flow branch separately and I still work on combining these two branches.
c. Related doc-string hasn't been updated. Plan to do it after b. done.

If any problem, please do not hesitate to tell me! Many thanks!

Kind regards,
Xianyuan

PR #35, opened 2021-01-08
Change `embed.video_cnn` to `embed.video_i3d` and `embed.video_res3d`.

### LINKED ISSUE #34: Unnecessary `kale.embed.video_cnn`
To me, `kale.embed.video_cnn` is not necessary just for two algorithms. This introduces unnecessary depth to the API. It will be needed only if `embed` becomes too big in future. Change it to something like `video_i3d` or `video_res3d` will be more compact and convenient to use.

XianyuanLiu:
Fixes #34.

### Description
a. Change `embed.video_cnn` to `embed.video_i3d` and `embed.video_res3d`.
b. Update the comments according to the #28.

### Status
Ready

### Types of changes


haipinglu (approved):
Looks good.

PR #44, opened 2021-01-20
isort all (imports)

haipinglu:
### Description
- Run `isort` on all files to sort the imports (skipped 4 files, changed 39 files, Flake8 checked).
- Switch on `isort` pre-commit and hope it can work well. Even if not, I should be able to fix it.
- I will try to fix some other pre-commits if feasible. 

These changes are light, affecting only the format of the imports so a quick check by you is much appreciate before merging. Thanks.

### Status
**Ready**

### Types of changes

### DIFF
--- a/examples/action_dann_lightn/main.py
+++ b/examples/action_dann_lightn/main.py
@@ -14,7 +14,7 @@ from examples.action_dann_lightn.config import get_cfg_defaults
 from examples.action_dann_lightn.model import get_model
 from kale.loaddata.multi_domain import MultiDomainDatasets
 from kale.loaddata.video_access import VideoDataset
-from kale.utils.csv_logger import setup_logger  # np error if move this to later, not sure why
+from kale.utils.csv_logger import setup_logger 
 from kale.utils.seed import set_seed
 
 # import sys
--- a/examples/action_dann_lightn/main.py
+++ b/examples/action_dann_lightn/main.py
@@ -7,19 +7,19 @@ import argparse
 import logging
 import os
 
+import pytorch_lightning as pl
+from pytorch_lightning import loggers as pl_loggers
+
+from examples.action_dann_lightn.config import get_cfg_defaults
+from examples.action_dann_lightn.model import get_model
+from kale.loaddata.multi_domain import MultiDomainDatasets
+from kale.loaddata.video_access import VideoDataset
+from kale.utils.csv_logger import setup_logger  # np error if move this to later, not sure why
+from kale.utils.seed import set_seed
+
 # import sys
 
 # sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "../..")))
-
-from kale.utils.csv_logger import setup_logger  # np error if move this to later, not sure why
-import pytorch_lightning as pl
-
-from pytorch_lightning import loggers as pl_loggers
-from examples.action_dann_lightn.config import get_cfg_defaults
-from examples.action_dann_lightn.model import get_model
-from kale.loaddata.video_access import VideoDataset
-from kale.loaddata.multi_domain import MultiDomainDatasets
-from kale.utils.seed import set_seed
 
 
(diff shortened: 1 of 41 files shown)


examples/digits_dann_lightn/main.py
```diff
@@ -3,24 +3,27 @@
...
 import argparse
+
 # import warnings
 # import sys
 import logging
+import os
 
-# sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '../..')))
-
-from kale.utils.csv_logger import setup_logger  # np error if move this to later, not sure why
```
XianyuanLiu, line 14: So this error doesn't exist after moving, does it?

examples/action_dann_lightn/main.py
```diff
@@ -7,20 +7,20 @@
 import logging
 import os
 
-# import sys
-
-# sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "../..")))
-
-from kale.utils.csv_logger import setup_logger  # np error if move this to later, not sure why
```
XianyuanLiu, line 14: So this error doesn't exist after moving, does it?

XianyuanLiu: Checked it. No error on Win10.

examples/digits_dann_lightn/main.py
XianyuanLiu: Checked. No error but a Numpy warning which seems not related to this change. That's fine.
`UserWarning: The given NumPy array is not writeable, and PyTorch does not support non-writeable tensors.`

XianyuanLiu (approved):
The sort is nice! No problem with computer vision codes. Many thank!

examples/digits_dann_lightn/main.py
haipinglu: > So this error doesn't exist after moving, does it?

The previous wrapping was due to length 88. This line is <120 so no problem after setting black to length 120. That being said, you raised a very good point. Now the imports are resorted. That kind of fix will not be allowed anymore. I will remove the comment. Glad to see you have tested that there is no error but just a numpy warning. Many thanks! Peer review really improves quality!

RaivoKoot (approved):

pz-white (approved):
LGTM for GraphML part, thanks for the import sort adjustment!

sz144 (approved):
The changes in mpca.py are okay.

examples/digits_dann_lightn/main.py
XianyuanLiu: Sorry to unresolve this, I find it is okay on my pc with Win10, but it raises an error on HPC. After reverting it, the error is solved.

```
Traceback (most recent call last):
  File "main.py", line 13, in <module>
    from examples.action_dann_lightn.config import get_cfg_defaults
ModuleNotFoundError: No module named 'examples'
All is done!
```

haipinglu: @XianyuanLiu in your examples, you could import those local modules as `from config import get_cfg_defaults`, as in digit_dann_lightn We can talk tomorrow if this is till an issue. Thanks.

XianyuanLiu: Solved by `pip install .`. Many thanks 💯 
FTR, [set & save environment variable in conda envs](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#saving-environment-variables).

haipinglu: What is your second line of reply on "set & save environment variable in conda envs." for? I can't get it without context?

XianyuanLiu: Sorry. It is a record. Before you told me to use `pip install .`, I have followed the link to add `pykale` root path to the environment variable in my conda env and solved the problem. It is not the best option but may be useful in some other situations.

PR #52, opened 2021-02-02
Quick Raivo fix mypi

RaivoKoot:
### Description
Includes fixes needed to merge Haiping's fix-mypy PR.
I could not push directly to the repo, so need to do PR. (Also forgot that I maybe could have pushed these commits to your fix-mypy PR. If there are problems, I can delete this PR and repeat and push to the fix-mypy PR. Let me know if you run into problems and I should do that)

I fixed the mypi errors (including those in Xianyuan's video_datasets.py file) and also deleted video_transformer.py and linformer.py because those are wrong and should not be used. 

Also removed `tqdm` from requirements.txt again because pytorch lightning installs it automatically and requires a different version which is okay.
### Status
Ready to merge

### Types of changes

- [x] Documentation updated, tested `make html` command in the `docs/` folder.


haipinglu:
Many thanks Raivo.

Yes, you should directly commit mypy-related changes to the mypy branch but NOT to merge yet please. I hope to check with Xianyuan about his module on possible fixes rather than ignore (we should try to fix and use ignore only when absolutely necessary). 

Will deleting video_transformer.py and linformer.py cause errors in other parts such as the cifar_cnntransformer example where I guess they were called (e.g. someone may run the example after merge)?

Also please commit the tqdm change to the dependencies branch directly. This will help avoid merging conflicts and also group related changes under one PR for easy review and more issue-focused management. Separate PRs for the same issue could be confusing in future. I also suggest during code review, the reviewer could fix simple, sure problems (such as typos) directly, which will be more efficient for the team than comment and fix.

As an example, this was what Xianyuan did for the print branch in PR #48. You have the same right as him so you can direct commit to a branch.

No one can push to the master directly, including me. This may be inconvenient but it is more robust.

haipinglu:
Also it is not clear whether the mypy checking has passed because it is not switched on here (only on in the mypy branch), though I believe it should pass.

bobturneruk:
Would a PR to merge `raivo_fix_mypi` into `hplu/fix-mypy` be useful here?

haipinglu:
> Would a PR to merge `raivo_fix_mypi` into `hplu/fix-mypy` be useful here?

@bobturneruk That means multiple PRs for the same issue and it feels complicated to me at least. Direct commit is much simpler in terms of history of a particular issue. Also I have other suggestions on the changes that should be committed to the dependencies branch (see my first comment).

@RaivoKoot What do you think? You can leave the video_dataset.py update out as I will check with @XianyuanLiu. We should use ignore as the last resort.

haipinglu (commented):
Pease see the comments above.

RaivoKoot:
> Yes, you should directly commit mypy-related changes to the mypy branch but NOT to merge yet please. 

Okay no problem. I will re-commit my changes into your branch.

> I hope to check with Xianyuan about his module on possible fixes rather than ignore (we should try to fix and use ignore only when absolutely necessary).  

Yes I agree. I investigated those 4 mypy errors and 2 of those were wrong. It was because mypy does not scan the imported library and thought the operation Xianyuan used was incorrect. However, the library he uses for the operation makes it possible.

> Will deleting video_transformer.py and linformer.py cause errors in other parts such as the cifar_cnntransformer example where I guess they were called (e.g. someone may run the example after merge)?

No. Currently nothing uses `video_transformer` or `linformer`. I pushed them to our api a while ago but never created a basic example with them and later on I decided they should not be used.  
  
Those two were somewhat separate from the initial `cnn_transformer` api and example, which I am leaving in the repo.


> Also please commit the tqdm change to the dependencies branch directly. This will help avoid merging conflicts and also group related changes under one PR for easy review and more issue-focused management. Separate PRs for the same issue could be confusing in future. I also suggest during code review, the reviewer could fix simple, sure problems (such as typos) directly, which will be more efficient for the team than comment and fix.
  
No problem. Will follow all of this

> @RaivoKoot What do you think?

Unfortunately, I don't know how to do that and am a little scared of messing it up. So, I will just re-commit to your PR.

PR #49, opened 2021-01-30
Set up mypy and add draft PR & US English guideline

haipinglu:
Fixes [mypy card](https://github.com/pykale/pykale/projects/9#card-53187365).

### Description
- `ignore_missing_imports` (MONAI, Kornia, scikit-learn, PyTorchLightning): reduces 77 errors in 37 files to 12 errors in 4 files
- Fix remaining 12 errors (to be done)
- Switch on mypy pre-commit
- Add [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) guideline
- Add US English spell check guideline

Key references
- [MONAI mypy config](https://github.com/Project-MONAI/MONAI/blob/master/setup.cfg)
- [Kornia mypy config](https://github.com/kornia/kornia/blob/master/setup.cfg)
- [Scikit learn mypy config](https://github.com/scikit-learn/scikit-learn/blob/main/setup.cfg)
- [Torchvision mypy config](https://github.com/pytorch/vision/blob/master/mypy.ini)
- [PyTorch Lightning mypy config](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/setup.cfg)

### Status
**Ready/Work in progress**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).

PR #53, opened 2021-02-04
Add DAN to digits_dann_lightn

XianyuanLiu:
### Description
1. Add DAN option in the `digits_dann_lightn/model.py` referring to [ADA](https://github.com/criteo-research/pytorch-ada/blob/ac924cc0780d0d914982950d05b19b4975752c9d/adalib/ada/utils/experimentation.py#L169).
2. Fix the bug raise by #49 in `kale/loaddata/multi_domain.py`. I reverted the changes and added `# type: dict` referring to [mypy issue](https://github.com/python/mypy/issues/922). Is it a good way?

### Status
**Ready**

### Types of changes

### DIFF
--- a/examples/digits_dann_lightn/configs/MN2UP-DAN.yaml
+++ b/examples/digits_dann_lightn/configs/MN2UP-DAN.yaml
@@ -1,0 +1,2 @@
+DAN:
+  METHOD: 'DAN'
--- a/examples/digits_dann_lightn/model.py
+++ b/examples/digits_dann_lightn/model.py
@@ -65,31 +65,43 @@ def get_model(cfg, dataset, num_channels):
     feature_dim = feature_network.output_size()
     classifier_network = ClassNetSmallImage(feature_dim, cfg.DATASET.NUM_CLASSES)
 
-    method = domain_adapter.Method(cfg.DAN.METHOD)
-    critic_input_size = feature_dim
-    # setup critic network
-    if method.is_cdan_method():
-        if cfg.DAN.USERANDOM:
-            critic_input_size = cfg.DAN.RANDOM_DIM
-        else:
-            critic_input_size = feature_dim * cfg.DATASET.NUM_CLASSES
-    critic_network = DomainNetSmallImage(critic_input_size)
-
     config_params = get_config(cfg)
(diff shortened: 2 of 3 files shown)


XianyuanLiu:
The reason why I push this is I find something wrong when I run DAN on `digits_dann_lightn` and `action_dann_lightn`. The MMD loss `T_mmd` has some values below 0. It will cause `T_total_loss` below 0 because `T_total_loss = T_task_loss + 1 * T_mmd`. Something is wrong, but I can't figure out. Sorry about that. I will raise an issue for this.

haipinglu:
On mypy: the solution mypy suggested is ` _source_by_split: Dict[<type>, <type>] = ..."` and we discussed before string and optional may be the correct types to fill in. I tried but probably used wrong python types so it was not successful. 

Your message on DAN is a bit confusing. Is it ready to merge or it needs further works to fix the problem.

I feel that branches are more convenient when collaborations are needed, saving lots of unnecessary branches or forks or merges.

We could discuss these after Monday's group meeting or in our next individual meeting. I have to get my teaching ready first.

XianyuanLiu:
1. I followed your suggestion to change the code (also follow [pytorch](https://github.com/pytorch/pytorch/blob/916af892b3e7bfa43cab877ee3f2fd7a50f576a8/torch/quantization/fx/quantize.py#L104)).
2. It is ready to merge. There is no DAN option in the `digits_dann_lightn` before. We can only use CDAN and DANN there. I followed the same way in [ADA](https://github.com/criteo-research/pytorch-ada/blob/ac924cc0780d0d914982950d05b19b4975752c9d/adalib/ada/utils/experimentation.py#L170) to add `create_mmd_based` to `model.py`. Thus, we can train and test DAN and JAN on `digits_dann_lightn` now. However, the loss problem is still existing. I push this because I hope others like Shuo can also run DAN in digits to test this problem.
3. I will create a new branch next time. 
4. Sure, we can discuss next Monday. No urgent and I can use CDAN.

haipinglu (approved):
Thank you for adding DAN to the digits example. That's correct. I did not implement or test DAN earlier. Glad you did now.

PR #76, opened 2021-03-15
Test mpca

sz144:
### Description
Test MPCA using gait gallery data. Test for uploading test data.

### Status
**Ready**

### Types of changes

- [ ] In-line docstrings updated and documentation `docs` updated.


haipinglu:
@bobturneruk The test workflow still cannot find the data file. Could you take a look to see what might be the cause?

bobturneruk:
I'll have a look.

tests/embed/test_mpca.py
```diff
@@ -0,0 +1,77 @@
+import numpy as np
+import pytest
+from numpy import testing
+from scipy.io import loadmat
+from tensorly.tenalg import multi_mode_dot
+
+from kale.embed.mpca import MPCA
+
+gait = loadmat("tests/test_data/gait_gallery_data.mat")
```
bobturneruk, line 9: Would this be better as a fixture @drj11, @robadob, @ptheywood, @davidwilby? I'm a bit uncomfortable with not passing it as an argument to a function...

bobturneruk (approved):
Without knowing exactly what the code to be tested is supposed to do (not being a domain expert), I'm bit limited here in what I can say. However, I've got a couple of suggestions:

- Some of the tests may be a bit redundant i.e. testing the number of dimensions and the size of each dimension of an array in three separate assert statements.
- Somewhat conversely, it might be better to split some of the tests down into several smaller tests. This can be useful when trying to quickly pin down exactly what has gone wrong - if any of the asserts fail in a test, the whole test fails.

These are both fairly subjective points and in my view the code is fine to merge as it is.

tests/embed/test_mpca.py
Robadob: Ideally you want a clean environment between tests, so one test doesn't infect the next. E.g. test a might break `gait`, making all subsequent tests fail.

If the cost of loading this stuff is huge (and would make running test suite impractical), could argue against that and just be very careful though.

A fixture would be a suitable way of automatically loading/deleting it per test, but I often get lazy and just copypaste my test setups between tests due to all the small per-test tweaks. Achieves the same thing, mostly a preference/standards thing.

bobturneruk: Thanks @robadob. I'll leave the decision to @sz144 in the context of my other comments.

sz144:
> 
> 
> Without knowing exactly what the code to be tested is supposed to do (not being a domain expert), I'm bit limited here in what I can say. However, I've got a couple of suggestions:
> 
>     * Some of the tests may be a bit redundant i.e. testing the number of dimensions and the size of each dimension of an array in three separate assert statements.
> 
>     * Somewhat conversely, it might be better to split some of the tests down into several smaller tests. This can be useful when trying to quickly pin down exactly what has gone wrong - if any of the asserts fail in a test, the whole test fails.
> 
> 
> These are both fairly subjective points and in my view the code is fine to merge as it is.

Thank you for your help and suggestions. I am also considering split the tests into smaller ones and hopefully create a new PR before our next meeting.

bobturneruk:
@sz144 - I'd advise getting this merged ASAP if @haipinglu is OK with that. Further improvements can then be made in subsequent PRs. PRs should be merged is they make the code better - don't wait until it's perfect! 😄

haipinglu (approved):

PR #82, opened 2021-03-19
Remove requirements in examples and update setup

haipinglu:
### Description
Remove requirements in examples and update setup

### Status
**Ready**

### Types of changes


bobturneruk (approved):
This seems to move all requirements for examples into the [extras] install, in-keeping with how [extras] is described in `readme.md` and simplifying the contents of the repository. I note that the `bindingdb_deepdta` example requires an additional install of `PyTDC` in it's readme. There may be other instances of this kind of thing (i.e. requirements sepcified in documentation) that could also be moved into the [extras] requirements at some point.

I can't test this without running all of the examples - and this would be a very involved process. Perhaps something to consider doing after this PR is merged?

haipinglu:
> This seems to move all requirements for examples into the [extras] install, in-keeping with how [extras] is described in `readme.md` and simplifying the contents of the repository. I note that the `bindingdb_deepdta` example requires an additional install of `PyTDC` in it's readme. There may be other instances of this kind of thing (i.e. requirements sepcified in documentation) that could also be moved into the [extras] requirements at some point.

Good catch. Sorry I overlooked that and will add `PyTDC` to [extra] and remove it from that example. [RDKit](https://github.com/rdkit/rdkit) seems to offer conda installation only for recent update (the [PyPI version](https://pypi.org/project/rdkit/) was last updated in 2009) so I did not find a good way to incorporate it in [extra].
I've checked other examples and did not find similar problems.
@pz-white Please take note of the change and in future updates.

> 
> I can't test this without running all of the examples - and this would be a very involved process. Perhaps something to consider doing after this PR is merged?

Yes, it will be good to test all examples, i.e. do the regression tests. However, at the moment, I think unit tests should be our priority to improve the test coverage of the core API from 3 to ~80 so that we can make an official release and get closer to join the PyTorch ecosystem.

PR #85, opened 2021-03-30
Test utils logger

bobturneruk:
### Description
Adds tests for and improvements to https://github.com/pykale/pykale/blob/master/kale/utils/logger.py.

### Status
**Work in progress**

- Select at least one appropriate label.

### Types of changes
- [x] In-line docstrings updated and documentation `docs` updated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,7 @@ examples/action_dann_lightn/configs_xianyuan/
 
 # Logs
 log-*.txt
-gitdiff.patch
+*gitdiff.patch
 
 # Sphinx documentation
 docs/build/
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ docs/source/backup
 
 # Testing
 .coverage
+cov.xml
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ examples/action_dann_lightn/configs_xianyuan/
 
 # Logs
 log-*.txt
+gitdiff.patch
 
 # Sphinx documentation
 docs/build/
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,9 @@ examples/*/lightning_logs/
 examples/*/demo*/
 examples/action_dann_lightn/configs_xianyuan/
 
+# Logs
+log-*.txt
+
 # Sphinx documentation
 docs/build/
 docs/source/backup
--- a/kale/utils/logger.py
+++ b/kale/utils/logger.py
@@ -3,20 +3,16 @@
 import datetime
 import logging
 import os
-import shlex
-import subprocess
+import uuid
 
 
-def git_hash():
-    """Gets a hash for different runs to have unique logfile names."""
-    cmd = 'git log -n 1 --pretty="%h"'
-    ret = subprocess.check_output(shlex.split(cmd)).strip()
-    if isinstance(ret, bytes):
-        ret = ret.decode()
-    return ret
+def log_file_name():
+    """Creates a log file name concatenating a formatted date and uuid"""
+    date = str(datetime.datetime.now().strftime("%m%d%H"))
+    return f"log-{date}-{str(uuid.uuid4())}.txt"
 
 
-def construct_logger(name, save_dir):
+def construct_logger(name, save_dir, file_name=log_file_name()):
     """Constructs a simple txt logger with a specified name at a specified path
 
     Reference: https://docs.python.org/3/library/logging.html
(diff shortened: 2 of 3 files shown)


kale/utils/logger.py
```diff
@@ -3,17 +3,17 @@
...
-        ret = ret.decode()
-    return ret
+def out_file_core():
+    """Creates an output file name concatenating a formatted date and uuid, but without an extension.
+
+    Returns:
+        string: A string to be used in a file name.
+    """
+    date = str(datetime.datetime.now().strftime("%Y%d%m_%H%M%S"))
+    return f"log-{date}-{str(uuid.uuid4())}"
```
willfurnass, line 16: Or more concisely: `f"log-{datetime.datetime.now():%Y%d%m_%H%M%S}-{uuid.uuid4()}"`

kale/utils/logger.py
```diff
@@ -3,17 +3,17 @@
...
 
 
 def construct_logger(name, save_dir):
```
willfurnass, line 19: Docstring?  It's not clear what lines 38/39 do.

tests/utils/test_logger.py
```diff
@@ -0,0 +1,57 @@
...
+import os
+
+import pytest
+
+from kale.utils import logger
+
+
+@pytest.fixture(scope="session")
+def testing_logger():
+    save_dir = "./"
```
willfurnass, line 11: `os.getcwd()`?

willfurnass (commented):
Made some minor suggestions but would need to read up on fixtures to be able to review the rest!

kale/utils/logger.py
bobturneruk: Thanks!

kale/utils/logger.py
bobturneruk: Thanks - I've changed that.

kale/utils/logger.py
bobturneruk: I've kept it verbose for now.

tests/utils/test_logger.py
bobturneruk: That's better, thanks.

.gitignore
```diff
@@ -24,9 +24,14 @@ examples/*/lightning_logs/
 examples/*/demo*/
 examples/action_dann_lightn/configs_xianyuan/
 
+# Logs
+log-*.txt
```
haipinglu, line 28: Another option is to do a clean up: https://fastai1.fast.ai/dev/test.html#after-test-cleanup
Up to you which is better.

bobturneruk: Test outputs are currently torn down after this line:

https://github.com/pykale/pykale/blob/e3d0329f34a952ac5b81514fe2829e075e244d31/tests/utils/test_logger.py#L13

I added logs to `.gitignore` for convenience when running the code normally. The tests should not leave files lying around.

haipinglu: I don't know that before but now I learned. Thanks. Is this the official doc? : https://docs.pytest.org/en/stable/fixture.html#yield-fixtures-recommended

tests/utils/test_logger.py
```diff
@@ -0,0 +1,57 @@
...
+import pytest
+
+from kale.utils import logger
+
+
+@pytest.fixture(scope="session")
+def testing_logger():
+    save_dir = os.getcwd()
+    testing_logger = logger.construct_logger("test", save_dir)
+    yield testing_logger
```
haipinglu, line 13: I see. So the above is run first to construct a logger. Then the test cases check the validity of the (type/existence of) resulted objects. After tests, the teardown cleans up the generated files.

Thanks for the solution. I learned a lot!

haipinglu (approved):

tests/utils/test_logger.py
```diff
@@ -0,0 +1,57 @@
+import logging
+import os
+
+import pytest
+
+from kale.utils import logger
+
+
+@pytest.fixture(scope="session")
```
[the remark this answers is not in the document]
bobturneruk: Yes. I think you may be right. I'll try this out.

haipinglu:
> While these tests "cover" the required code, are there other things I should be checking?

It looks fine to me, though I am a newbie myself.

> Is the use of fixtures appropriate?

It looks fine. I made only one comment about the scope, "module" should be the default unless it is global (conftest). I mainly refer to fastai1 (lots of examples), e.g. https://github.com/fastai/fastai1/blob/master/tests/test_basic_train.py

I think it can be merged unless you want to add another.

tests/utils/test_logger.py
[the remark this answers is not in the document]
haipinglu: > Yes. I think you may be right. I'll try this out.

The five scopes (default: "function") are explained at https://docs.pytest.org/en/stable/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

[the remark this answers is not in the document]
bobturneruk: I think it's good practise to use the smallest scope that works.

bobturneruk:
I think this is good to merge (if all the checks pass). I'm not authorised to merge, it seems.

haipinglu:
@bobturneruk The problem has been solved with the help of Xianyuan. It is the "Restrict who can push to matching branches" setting for protected branch. If ticked, I need to manually add you all.

So this merge was done by Xianyuan rather than me.

PR #88, opened 2021-04-08
Automatically sort documented members by source order

pz-white:
### Description
Change doc config file and modify the default sorted method from alphabetical to source order.
### Status
Ready

### Types of changes

- [x] In-line docstrings updated and documentation `docs` updated.


pz-white:
@haipinglu I find a new solution to automatically sort doc members by sorce order and do not need to modify `rst` files one by one. The Sphinx supports change default sort method by setting in conf file. Test on local and work well.

haipinglu (approved):

PR #90, opened 2021-04-08
Test mpca

sz144:
### Description
Minor update for mpca test.

### Status
**Ready**

### Types of changes

- [ ] In-line docstrings updated and documentation `docs` updated.

### DIFF
--- a/tests/embed/test_mpca.py
+++ b/tests/embed/test_mpca.py
@@ -11,6 +11,7 @@ baseline_model = loadmat("tests/test_data/mpca_baseline_res.mat")
 
 N_COMPS = [1, 50, 100]
 VAR_RATIOS = [0.7, 0.95]
+relative_tol = 0.00001
 
 
 @pytest.mark.parametrize("n_components", N_COMPS)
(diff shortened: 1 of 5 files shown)


tests/embed/test_mpca.py
```diff
@@ -47,7 +47,7 @@ def test_mpca(var_ratio, n_components):
     testing.assert_equal(x0_proj.ndim, 2)
     testing.assert_equal(x0_proj.shape[0], 1)
     testing.assert_equal(x0_proj.shape[1], n_components)
-    x0_rec = mpca.inverse_transform(x0_proj)
+    x0_rec = mpca.inverse_transform(x0_proj.reshape(-1))
```
haipinglu, line 50: A bit difficult to review. Can you explain a bit why the change?

tests/embed/test_mpca.py
```diff
@@ -59,19 +59,16 @@ def test_mpca(var_ratio, n_components):
...
+    baseline_mean = baseline_model['TXmean']
     mpca = MPCA(var_ratio=0.97)
     x_proj = mpca.fit(x).transform(x)
-    baseline_proj_x = multi_mode_dot(x, baseline_proj_mats, modes=[1, 2, 3])
-    # check whether the output shape is consistent with the baseline output by keeping the same variance ratio 97%
-    testing.assert_equal(x_proj.shape, baseline_proj_x.shape)
+    testing.assert_allclose(baseline_mean, mpca.mean_)
+    baseline_proj_x = multi_mode_dot(x - baseline_mean, baseline_proj_mats, modes=[1, 2, 3])
+    # check whether the output embeddings is close to the baseline output by keeping the same variance ratio 97%
+    testing.assert_allclose(x_proj ** 2, baseline_proj_x ** 2, rtol=0.00001)
```
haipinglu, line 68: The tolerance value is used more than once so it may be better to set it as a "global" variable at the top (and in future, maybe in `conftest.py`)

tests/embed/test_mpca.py
```diff
@@ -59,19 +59,16 @@ def test_mpca(var_ratio, n_components):
...
-        for j in range(baseline_proj_mats[i].shape[0]):
-            # subtraction of eigen-vector columns
-            eig_col_sub = mpca.proj_mats[i][j:] - baseline_proj_mats[i][j:]
-            # sum of eigen-vector columns
-            eig_col_sum = mpca.proj_mats[i][j:] + baseline_proj_mats[i][j:]
-            compare_ = np.multiply(eig_col_sub, eig_col_sum)
-            # plus one for avoiding inf relative difference
-            testing.assert_allclose(compare_ + 1, np.ones(compare_.shape))
+        # testing.assert_allclose(abs(mpca.proj_mats[i]), abs(baseline_proj_mats[i]))
+        testing.assert_allclose(mpca.proj_mats[i] ** 2, baseline_proj_mats[i] ** 2, rtol=0.00001)
```
haipinglu, line 74: See above regarding `rtol` value.

tests/embed/test_mpca.py
sz144: will change

tests/embed/test_mpca.py
sz144: Before: the shape of `x0_proj` is (1, n_features). After: the shape of `x0_proj` is (n_features,). Here we test whether a vector can be reconstructed as tensor in original shape. It is also correct. I made the change here simply to cover [line 249](https://github.com/pykale/pykale/blob/master/kale/embed/mpca.py#L249) of mpca.py.

haipinglu (approved):

PR #101, opened 2021-04-18
Keep __init__ simple without public and private interfaces

### LINKED ISSUE #56: Public and Internal Interfaces
This is from a card created by @sz144 

References:
https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces

https://stackoverflow.com/questions/35727134/module-imports-and-init-py-in-python

Examples:
- [pytorch](https://github.com/pytorch/pytorch/blob/master/torch/__init__.py)
- [scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/decomposition/__init__.py)
- [tensorly](https://github.com/tensorly/tensorly/blob/main/tensorly/tenalg/__init__.py)

haipinglu:
Fixes #56

### Description
I have studied the issues concerning the said public and private interfaces raised by Shuo earlier. My conclusion is that at the current state of our development, I do not see implementing public/private interfaces beneficial to us. Here, we take Shuo's original commented code in [`kale\embed\__init__.py`](https://github.com/sz144/pykale/blob/mpca/kale/embed/__init__.py) as an example (I purposely used Shuo's forked branch since the code here will be removed), consisting of two parts: 
1. Line 1 to 26: imports of various submodules of this module 
2. Line 28 to 40: the definition of `__all__` of all public submodules

I did not see the benefits but instead found problems with such practices considering the following: 
- `__all__` (line 28-40) is for wildcard import of the form `from ... import *`, which is [considered as a bad practice by the official python documentation](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package). In what scenario, will we want to import all the submodules in `kale.embed`? So far, we did not use it and I think we should NOT use it in future unless there are VERY strong reasons to do so. It will be the best to _explicitly import the submodules needed_. Therefore, defining `__all__` seems no longer useful.
- Code in `__init__` [will be executed](https://docs.python.org/3/tutorial/modules.html#packages) so those listed submodules (line 1 to 26) will be imported. But looking at the code, do we really need to import so many modules in a typical usage of pykale? I never see such a usage yet. These unnecessary imports will slow things down and are _against our principle of minimalism_. Keeping `__init__` simple is in line with the design principles of pykale.
- It takes time to write such interfaces but I do not see what benefits they bring. This is not the best of our limited time, considering other priority such as testing. I understand these interfaces are seen in many top packages, but that sole reason is not convincing to devote our efforts while considering the above problems, _unless there are stronger reasons that I am not aware of, which I will be glad to hear if you know._

Based on the above, while my original plan for this branch (and PR) was to implement public/private interfaces, eventually, what I did is to create and tidy up the `__init__` file of our modules. 

### Status
**Ready**

### Types of changes


pz-white (approved):

XianyuanLiu (approved):
When I don't define `i3d_joint` in `__all__` but import it, it shows a warning as below by PyCharm. It may prevent misuse and show the interfaces clearly. 

However, I haven't found some official documents to encourage people to use it, and the price of doing this for all kale files is too high. Thus, I agree with you. There is no very strong reason to do so. Thanks.

image

RaivoKoot:
Agreed!

RaivoKoot (approved):

haipinglu:
> When I don't define `i3d_joint` in `__all__` but import it, it shows a warning as below by PyCharm. It may prevent misuse and show the interfaces clearly.

My suggestion is not to use `__all__` at all. If you do not define `__all__`, there won't be such a warning. 

It may be better to describe the interfaces in docstrings to that they will show up in the documentation, rather than define them in code via `__all__`, which becomes cumbersome and unpleasant to read when there are many such as line 28-40 of  [`kale\embed\__init__.py`](https://github.com/sz144/pykale/blob/mpca/kale/embed/__init__.py).

As for misuse, we can start with an underscore to designate an internal function.

> 
> However, I haven't found some official documents to encourage people to use it, and the price of doing this for all kale files is too high. Thus, I agree with you. There is no very strong reason to do so. Thanks.
> 
> image

sz144 (approved):

haipinglu:
@bobturneruk I will merge this PR. If you have any comment, please let us know.

bobturneruk:
I approve the merge. While PEP8 says we should use `__all__`, in practise this doesn't always happen. I think what we are now doing with using prefix `_` to denote "private" is an appropriate balance between the time required to produce a more tightly controlled public interface, and user needs.

haipinglu:
> I approve the merge. While PEP8 says we should use `__all__`, in practise this doesn't always happen. I think what we are now doing with using prefix `_` to denote "private" is an appropriate balance between the time required to produce a more tightly controlled public interface, and user needs.

Thanks Bob. That's helpful to hear.

PR #113, opened 2021-04-22
Update reference to pykale master to pykale main

haipinglu:
### Description
Update reference to pykale master to pykale main after we renamed our default branch to main.

### Status
**Ready**

### Types of changes


bobturneruk (approved):
Looks good. Does readthedocs need an update at all?

haipinglu:
> Looks good. Does readthedocs need an update at all?

Yes, you are right. Thanks for the reminder. I just did :-)

PR #122, opened 2021-04-25
Test deep_dta

pz-white:
### Description
1. Add tests covering deepdta related modules including:
- kale/embed/seq_nn.py
- kale/loaddata/tdc_datasets.py
- kale/pipeline/deep_dti.py
- kale/predict/decode.py
- kale/prepdata/chem_transform.py

2. Introduce conda virtual environment in test action.

3. Remove MNIST/MNISTM test as they are blocked due to HTTPServer error.

### Status
**Ready**

### Types of changes


pytest.ini
```diff
@@ -0,0 +1,6 @@
+# pytest.ini
```
haipinglu, line 1: This is a good direction to resolve the challenge but we prefer to keep the number of files small so this should be configured in `pyproject.toml` instead. See [pytest docs on config using `pyproject.toml`](https://docs.pytest.org/en/stable/customize.html#pyproject-toml).

pz-white:
Reproduce the url error for MINIST, try to re-run after the server is stable.

haipinglu:
> Reproduce the url error for MINIST, try to re-run after the server is stable.

That's annoying. Hope it is temporary, otherwise, we will consider to resolve or skip it (MNIST/M) later.

bobturneruk:
Introduces `rdkit` dependency not installed by `pip install -e .[dev]`.

bobturneruk:
@pz-white can you run these tests locally in a clean `conda` environment using `pytest`?

pz-white:
@bobturneruk yes I can successfully run these test with pytest locally with my conda environment

.github/workflows/test.yml
```diff
@@ -32,9 +28,20 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+      - uses: s-weigand/setup-conda@v1
```
bobturneruk, line 31: If we need conda, consider https://github.com/conda-incubator/setup-miniconda

bobturneruk:
Seems like this PR introduces a lot of new requirements and changes CI quite a bit - we need to decide on how these will be managed. My default suggestion is to add to `setup.py` where possible, although I appreciate that `rdkit` needs `conda`. I suppose this means that `PyKale` now also requires `conda` unless the required functionality in `rdkit` can be installed some other way?

.github/workflows/test.yml
pz-white: Thanks Bob! Let me try it now.

haipinglu:
> Seems like this PR introduces a lot of new requirements and changes CI quite a bit - we need to decide on how these will be managed. My default suggestion is to add to `setup.py` where possible, although I appreciate that `rdkit` needs `conda`. I suppose this means that `PyKale` now also requires `conda` unless the required functionality in `rdkit` can be installed some other way?

Not sure how to do it. The official instruction to install `pytorch geometric` is bit different from regular packages, as in the updated `test.yml`. Do you think this can be put in `setup.py`? If yes, can you make the change? I don't know how yet.

.github/workflows/test.yml
pz-white: The conda should be necessary as I checked that the `rdkit` is not a pure python package so that the package author said it is not supported to distribute with `PyPi`

bobturneruk:
I'll have a look. I think this may impact on `PyKale` install instructions. I hope it doesn't mean extra manual steps.

bobturneruk:
I think we need to make it clearer what needs to be installed to run the tests: https://pykale.readthedocs.io/en/latest/installation.html

haipinglu:
> I think we need to make it clearer what needs to be installed to run the tests: https://pykale.readthedocs.io/en/latest/installation.html

Our current suggestion is pytorch and pytorch geometric should be manually installed, which is the most common practice because this often needs to match the hardware (cpu/gpu versions). 

Only those who works on graphs and use related APIs need to install pytorch geometric, not everybody (among 6 of us, only Peizhen is working on graphs. ). We have MSc students followed the instructions and it seems all right.

tests/embed/test_seq_nn.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def test_gcn_encoder():
+    gcn_encoder = GCNEncoder(in_channel=8, out_channel=32).eval()
+    assert gcn_encoder.conv1.__repr__() == "GCNConv(8, 8)"
+    assert gcn_encoder.conv2.__repr__() == "GCNConv(8, 16)"
+    assert gcn_encoder.conv3.__repr__() == "GCNConv(16, 32)"
+    N1, N2 = 4, 5
+    x = torch.randn(N1 + N2, 8)
+    batch = torch.tensor([0 for _ in range(N1)] + [1 for _ in range(N2)])
```
haipinglu, line 38: A bit hard to understand for me here, to learn from you later, maybe just because I am not familiar. If complicated, add some light comments may help others.

haipinglu (approved):
Good job!
@bobturneruk Shall we wait for your review before merging? Please let us know. Thanks.

bobturneruk:
Please don't wait. If there's anything specific you want me to look at retrospectively, let me know.

.github/workflows/test.yml
```diff
@@ -32,17 +28,28 @@ jobs:
...
         run: |
           pytest --cov=kale
-        shell: bash
+        shell: bash -l {0}
       - name: Report coverage
         run: |
           coverage xml
           bash <(curl -s https://codecov.io/bash)
-        shell: bash
+        shell: bash -l {0}
```
haipinglu, line 55: @pz-white May I know why you changed `bash` to `bash -l {0}`? Now [codecov](https://codecov.io/gh/pykale/pykale/tree/master/kale) is not updating and I am suspecting this to be the cause.

PR #123, opened 2021-04-26
Merge prep_cmr with image_transform plus tests

sz144:
Fixes [card](https://github.com/pykale/pykale/projects/9#card-55274306).

### Description

- Merging functions in `kale.prepdata.prep_cmr` with `kale.prepdata.image_transform`. 
- Adding tests for image transform.

### Status
**Ready**

### Types of changes


tests/prepdata/test_image_transform.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_reg():
+    n_samples = images.shape[-1]
+    coords = np.ones((n_samples, 4))
+    coords[:, 2:] += 20
+    coords[1:, :] += np.random.random(size=(n_samples - 1, 4))
+    with pytest.raises(Exception):
+        reg_img_stack(images, coords[1:, :])
+    images_reg, max_dist = reg_img_stack(images, coords)
+    testing.assert_allclose(images_reg, images)
```
haipinglu, line 21: I did not get fully your logic here and can only guess. How did you make the registered images to be close to the original? I though you randomly perturbed the coordinates of the n-1 (9) images. After registration, they are close. Is it because the random noise is of small value compared to the coords? You can explain to me with a voice message in WeChat or Skype. Thanks.

kale/prepdata/image_transform.py
```diff
@@ -81,3 +90,89 @@ def get_transform(kind, augment=False):
...
+        raise ValueError(error_msg)
+    n_landmarks = int(coords.shape[1] / 2)
+    dst_coord = coords[dst_id, :]
+    dst_coord = dst_coord.reshape((n_landmarks, 2))
+    max_dist = np.zeros(n_samples)
+    for i in range(n_samples):
+        if i == dst_id:
+            continue
+        else:
+            # epts = landmarks.iloc[i, 1:].values
```
haipinglu, line 120: Still useful? Otherwise, remove.

tests/prepdata/test_image_transform.py
```diff
@@ -0,0 +1,43 @@
+import numpy as np
+import pytest
+from numpy import testing
+from scipy.io import loadmat
+
+from kale.prepdata.image_transform import mask_img_stack, reg_img_stack, rescale_img_stack
+
+gait = loadmat("tests/test_data/gait_gallery_data.mat")
```
haipinglu, line 8: Smart to use gait to test image stacks. Good _reuse_.

kale/prepdata/image_transform.py
sz144: will remove it

tests/prepdata/test_image_transform.py
sz144: Sorry for the poor readability. Yes, the random noise to source coordinates are in (0, 1), which are small values compared to the image size. So the images after registration should be close to the original ones. The values in list `max_dist` are very small. Maybe we can test whether `max_dist + 1` is close to a ones vector (`+1` for avoiding inf relative difference). Because we do not have real images and landmark coordinates for testing here, this is probably the easiest way for me to test the function. I will add more comments to improve the readability.

haipinglu: Thanks for the clarification. That's helpful! Ready to merge.

haipinglu (approved):

PR #126, opened 2021-04-28
Re-add MNIST into testing

pz-white:
### Description
Re-add MNIST into testing

### Status
**Work in progress**

### Types of changes


pz-white:
It seems to some conflicts between torchvision and PyG dependency, close the PR and try to fix after converting to DGL.

PR #132, opened 2021-04-29
Create download.py

XianyuanLiu:
Fixes #128 option 2.

### Description
Create two data downloading functions for files and compressed files via GitHub URL. It hasn't cover @RaivoKoot 's one, which downloads files from Google drive.

### Status
**Ready**

### Types of changes


kale/utils/download.py
```diff
@@ -0,0 +1,71 @@
+# =============================================================================
+# Author: Xianyuan Liu, xianyuan.liu@sheffield.ac.uk
```
haipinglu, line 2: You may consider to start using your more permanent, personal email (or both emails, see my case at https://github.com/pykale/pykale/blob/main/kale/embed/mpca.py) as your Shef email will become invalid soon after you leave.

kale/utils/download.py
```diff
@@ -0,0 +1,71 @@
...
+#         Raivo Koot, rekoot1@sheffield.ac.uk
+#         Haiping Lu, h.lu@sheffield.ac.uk or hplu@ieee.org
+# =============================================================================
+
+"""Data downloading and compressed data extraction functions, Based on
+https://github.com/pytorch/vision/blob/master/torchvision/datasets/utils.py
+https://github.com/pytorch/pytorch/blob/master/torch/hub.py
```
haipinglu, line 9: I assume that you have followed our 3R green ML principles here to reuse pytorch APIs.

kale/utils/download.py
```diff
@@ -0,0 +1,71 @@
...
+        logging.info("Skipping Download and Extraction")
+        return
+    if not os.path.exists(output_directory):
+        os.makedirs(output_directory)
+    logging.info("Downloading and extracting {}.".format(output_file_name))
+    download_and_extract_archive(url=url, download_root=output_directory, filename=output_file_name)
+    logging.info("Datasets downloaded and extracted in {}".format(file))
+
+
+def download_file_by_url(url, output_directory, output_file_name):
```
haipinglu, line 47: _Reduce_: Quite some repetitions here. I suggest to merge `download_compressed_file_by_url` with this one by either
- add a flag to indicate whether it is compressed, or even better
- auto detect whether the file is in compressed format to switch to either `download_and_extract_archive` or `download_url_to_file`

kale/utils/download.py
haipinglu: However, it should be clear why the original APIs in pytorch is not good/simple enough. Why the user should use this API rather than the pytorch ones directly. What are the added benefits of using this API.

haipinglu (commented):
Our pre-commit hooks have been tested for long and are quite stable so I strongly suggest you to use the automation rather than _manual Run pre-commit auto format_ in your commits to improve efficiency, which is the professional practice and also has a _cleaner commit history_.

You should consider to have _tests_ written at the same time. This will save the overall coding time because you are working on the same functionality and also save review time because the tests help to validate the code as well.

See the comment at the top. Clarify the benefits against using pytorch APIs directly.

haipinglu:
Also, this PR did not successfully complete the auto project assignment action (https://github.com/pykale/pykale/runs/2471407104?check_suite_focus=true), saying "Resource not accessible by integration".

This is not critical but weird. In this case, we need to manually assign the project (please do). Not sure whether this is because the PR is from a fork. There were no problems before this PR.

You do not need to solve the problem (except adding project manually). It is just my observation and we may see how to solve it later. I will add a card.

haipinglu:
On `pre-commit`, it is also weird on your system because, if your pre-commit did not pass, you should not be able to make a commit so it is not clear why a `Run pre-commit auto format` is there. This somehow implies that your pre-commit hooks are not functioning automatically before each commit on your local system. I think this should be fixed, with help from Shuo maybe.

On my system, I cannot make a commit before pass the pre-commit checks, I believe this is the case for all other members, except you, so your commit history is noisier and unnecessarily long.

haipinglu:
> On `pre-commit`, it is also weird on your system because, if your pre-commit did not pass, you should not be able to make a commit so it is not clear why a `Run pre-commit auto format` is there. This somehow implies that your pre-commit hooks are not functioning automatically before each commit on your local system. I think this should be fixed, with help from Shuo maybe.
> 
> On my system, I cannot make a commit before pass the pre-commit checks, I believe this is the case for all other members, except you, so your commit history is noisier and unnecessarily long.

You can consider VS code for committing to GitHub if PyCharm problems cannot be fixed, Shuo used both. I am experienced in VS code.

kale/utils/download.py
XianyuanLiu: 1. `download_and_extract_archive` and `download_url_to_file` are not in the same PyTorch file and It would be better to integrate them into one file. This file cover functions to download files with different types from different sources. Using the same format and parameters like `output_directory` and `output_file_name` can also simplify our coding.

2. PyTorch provides the basic download function but it may raise issues when the `output_directory` does not exist. I only meet this problem.

haipinglu:
You have tested download in the other PR so that's why I think this separation just increased the amount of work.

haipinglu:
> You have tested download in the other PR so that's why I think this separation just increased the amount of work, without much added value. This is also why I encourage you to think about the workflow and the efficiency of your work.

https://github.com/pykale/pykale/blob/test-video_access/tests/loaddata/test_video_access.py

XianyuanLiu:
> But as said earlier, let us try to work on _direct branches rather than forks_.

Sorry about that. I create a new branch via Desktop and the default one is forked. I will check and use the direct one.

> You can consider VS code for committing to GitHub if PyCharm problems cannot be fixed, Shuo used both. I am experienced in VS code.

I will solve the problem. I commit and push via Desktop/PyCharm with automatic pre-commit. In the first commit, `isort` always raise issues and change the imports to the correct one. But when I commit again, the code passes the check but revert the `isort` first change. It is fine when I use Terminal to run `pre-commit run --all` and commit.

haipinglu (approved):
Great. This will greatly simplify our related coding!

PR #135, opened 2021-05-02
Update docs, docs CI, and example readmes

haipinglu:
### Description
- Update (outdated) readmes for two examples `cifar_isonet` and `digits_dann_lightn`
- Update docs requirements to fix missing docs
- Add pytorch tests to tests readme
- Use RTD docs CI instead
### Status
**Ready**

### Types of changes


docs/source/kale.embed.rst
```diff
@@ -30,7 +29,6 @@ kale.embed.gcn module
    :members:
    :undoc-members:
    :show-inheritance:
-   :exclude-members: message, update
```
pz-white, line 33: May I ask why the exclude-members are deleted? In my understanding the message and update should be private methods and not be emerged in docs. I follow the rules on pytorch-geometrics doc: https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.conv.GCNConv

docs/source/kale.prepdata.rst
```diff
@@ -31,7 +31,7 @@ kale.prepdata.tensor\_reshape module
    :undoc-members:
    :show-inheritance:
 
-kale.prepdata.chem\_transform module
+kale.prepdata.chem_transform module
```
pz-white, line 34: Does this change work on doc build? As I remeber it failed without "\" on my local build.

docs/source/kale.utils.rst
```diff
@@ -6,7 +6,7 @@ Utilities
 Submodules
 ----------
 
-kale.utils.csv\_logger module
+kale.utils.csv_logger module
```
pz-white, line 9: Same as above.

docs/source/kale.embed.rst
haipinglu: I have thought it was the cause of the missing docs. I proved that I was wrong. It was due to the libraries. I will reverse and test.

docs/source/kale.prepdata.rst
haipinglu: It seems work both ways on rtd. I made the changes trying to find out why docs are missing on rtd

pz-white (approved):
All tests are passed! This PR LGTM, thanks Haiping!

PR #136, opened 2021-05-02
Test domain adapter

haipinglu:
### Description
- Add tests for domain adapter
- WDGRLMod: fix type error in `optimizer_step` but not fully covered
- Use USPS (23MB) now so the full test takes 20m or so on server. Later construct smaller random datasets or use small subsets of real data to replace USPS

### Status
**Ready**

### Types of changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,6 @@ docs/source/backup
 .coverage
 cov.xml
 tests/test_data/download
+
+# Lightning
+lightning_logs
--- a/kale/pipeline/domain_adapter.py
+++ b/kale/pipeline/domain_adapter.py
@@ -895,6 +895,7 @@ class WDGRLtrainerMod(WDGRLtrainer):
             # "log": log_metrics,
         }
 
+    # Add on_tpu=False etc following https://github.com/PyTorchLightning/pytorch-lightning/issues/2934 to fix error for WDGRLMod: TypeError: optimizer_step() got an unexpected keyword argument 'on_tpu'
     def optimizer_step(
         self,
         current_epoch,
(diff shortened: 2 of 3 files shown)


XianyuanLiu (approved):
Thanks. Looks good to me.

We may add more detailed tests _in the future_ for `sampler` and `losses`. The current one may be too high-level. We can create the simplest model and data loaders like [`boring_model`](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/helpers/boring_model.py#L63) to fix the loss output for the test. We can also test our sampler output size by following [PyTorch sampler test](https://github.com/pytorch/vision/blob/master/test/test_datasets_samplers.py).

TorchVision test: https://github.com/pytorch/vision/tree/master/test
PyTorch-Lightning test:  https://github.com/PyTorchLightning/pytorch-lightning/tree/master/tests

haipinglu:
> Thanks. Looks good to me.
> 
> We may add more detailed tests _in the future_ for `sampler` and `losses`. The current one may be too high-level. We can create the simplest model and data loaders like [`boring_model`](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/helpers/boring_model.py#L63) to fix the loss output for the test. We can also test our sampler output size by following [PyTorch sampler test](https://github.com/pytorch/vision/blob/master/test/test_datasets_samplers.py).
> 
> TorchVision test: https://github.com/pytorch/vision/tree/master/test
> PyTorch-Lightning test: https://github.com/PyTorchLightning/pytorch-lightning/tree/master/tests

Good suggestion and links. I've put a new card in 0.2.0 https://github.com/pykale/pykale/projects/10#card-60306281

Issue #148, opened 2021-05-13
[Bug] `digits_dann_lightn` example command from readme returns an error as it tries to download data from a server.


bobturneruk:
# 🐛 Bug

`digits_dann_lightn` example command from readme returns an error as it tries to download data from a server.

## To reproduce

```shell
python main.py --cfg configs/MN2UP-CDAN.yaml --gpus 0
```

** Stack trace/error message **
```
DAN:
  METHOD: CDAN
  RANDOM_DIM: 1024
  USERANDOM: False
DATASET:
  DIMENSION: 784
  NAME: digits
  NUM_CLASSES: 10
  NUM_REPEAT: 10
  ROOT: ../data
  SIZE_TYPE: source
  SOURCE: mnist
  TARGET: usps
  WEIGHT_TYPE: natural
OUTPUT:
  DIR: ./outputs\digits_mnist2usps
  PB_FRESH: 0
  ROOT: ./outputs
  VERBOSE: False
SOLVER:
  AD_LAMBDA: True
  AD_LR: True
  BASE_LR: 0.001
  INIT_LAMBDA: 1
  MAX_EPOCHS: 120
  MIN_EPOCHS: 20
  MOMENTUM: 0.9
  NESTEROV: True
  SEED: 2020
  TEST_BATCH_SIZE: 200
  TRAIN_BATCH_SIZE: 150
  TYPE: SGD
  WEIGHT_DECAY: 0.0005
==> Building model for seed 2020 ......
Downloading http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz to ../data\MNIST\raw\train-images-idx3-ubyte.gz
0it [00:00, ?it/s]Traceback (most recent call last):
  File "main.py", line 91, in <module>
    main()
  File "main.py", line 59, in main
    model, train_params = get_model(cfg, dataset, num_channels)
  File "C:\Users\bobturner\Documents\pykale\examples\digits_dann_lightn\model.py", line 98, in get_model
    model = domain_adapter.create_dann_like(
  File "c:\users\bobturner\documents\pykale\kale\pipeline\domain_adapter.py", line 149, in create_dann_like
    return CDANtrainer(
  File "c:\users\bobturner\documents\pykale\kale\pipeline\domain_adapter.py", line 602, in __init__
    super().__init__(dataset, feature_extractor, task_classifier, critic, **base_params)
  File "c:\users\bobturner\documents\pykale\kale\pipeline\domain_adapter.py", line 468, in __init__
[...]

PR #149, opened 2021-05-17
Digits notebook

bobturneruk:
Associated with https://github.com/pykale/pykale/discussions/147

### Description
Idea is to test the feasibility of adding an interactive notebook e.g. with myBinder or Google Collab (or both).

### Status
**Work in progress**

### Types of changes


bobturneruk:
Currently invoked with:

```
python main.py --cfg configs/QUICK-AND-DIRTY.yaml
```

haipinglu:
> It's taking a very long time to run (>half an hour) even with a small number of epochs.

To reduce further
- Use MAX_EPOCHS=2 (reduce 1/3)
- Use just 2 out of 10 classes, e.g. 1 and 2.  Need to write code to take subsets of ` source` and `target` (by class label).

Possible to do a profiling to understand the most costly part? We may brainstorm more ideas on Thursday too.

bobturneruk:
Profiling indicates this takes about an hour to run on my PC as is, so I'll look at reducing number of classes.

bobturneruk:
Is the envisioned approach to extract train and test sets from source and target, subset, then rebuild the datasetaccess objects for subsequent use e.g. at https://github.com/pykale/pykale/blob/8177ad9a79b3f6d96b3538a34a67514cf1a82b0b/examples/digits_dann_lightn/main.py#L48
Maybe I'm misunderstanding or not explaining myself well!

haipinglu:
> Is the envisioned approach to extract train and test sets from source and target, subset, then rebuild the datasetaccess objects for subsequent use e.g. at
> 
> https://github.com/pykale/pykale/blob/8177ad9a79b3f6d96b3538a34a67514cf1a82b0b/examples/digits_dann_lightn/main.py#L48

Yes, the line above get the `source` and `target` datasets and this line do the splitting to construct a source-target domain dataset (hence multidomain).

Oh, this question helps me spot another way to reduce the computational cost, i.e. setting `val_split_ratio` to a higher ratio. Default is `0.1` so using 90% for training. We can set it to a higher number, e.g. `0.5` so that using only 50% samples for training, which will give lower accuracy but save quite some training time.

examples/digits_dann_lightn/main.py
```diff
@@ -46,7 +46,11 @@ def main():
         DigitDataset(cfg.DATASET.SOURCE.upper()), DigitDataset(cfg.DATASET.TARGET.upper()), cfg.DATASET.ROOT
     )
     dataset = MultiDomainDatasets(
-        source, target, config_weight_type=cfg.DATASET.WEIGHT_TYPE, config_size_type=cfg.DATASET.SIZE_TYPE
+        source,
+        target,
+        config_weight_type=cfg.DATASET.WEIGHT_TYPE,
+        config_size_type=cfg.DATASET.SIZE_TYPE,
+        val_split_ratio=0.5,
```
haipinglu, line 53: Better to change it via yaml config rather than hardcode, e.g., cfg.DATASET.VAL_RATIO.

haipinglu: To get it running first, you can then config to cfg.DATASET.VAL_RATIO=0.9 in the fast version, but still keeping the default 0.1 as the default in config.py.

bobturneruk: Sure. I was confused by it having a different name.

bobturneruk: Would `DATASET.NUM_REPEAT` be another target to reduce?

haipinglu: > Sure. I was confused by it having a different name.

What does your "it" refer to? VAL_SPLIT_RATIO? I was hoping to make it more compact but if that causes confusion, just use cfg.DATASET.VAL_SPLIT_RATIO

haipinglu: > Would `DATASET.NUM_REPEAT` be another target to reduce?

Oh, I forgot mention that. That's right and we are almost there!
That should be the first to reduce. Just change it to **1** and you will get **10x** improvement. No need to repeat for notebook. Then the notebook should only take minutes to run for 10 classes, and maybe 1 min for 2 classes.

If you think flexible subset is good, we can also make the number of classes in subset a configurable variable.

bobturneruk:
Got it down to about 5 minutes on my machine. I'll put an notebook together and see how long it takes in myBinder without further modification.
```
Days              : 0
Hours             : 0
Minutes           : 4
Seconds           : 56
Milliseconds      : 189
Ticks             : 2961896145
TotalDays         : 0.00342812053819444
TotalHours        : 0.0822748929166667
TotalMinutes      : 4.936493575
TotalSeconds      : 296.1896145
TotalMilliseconds : 296189.6145
```

bobturneruk:
WIP myBinder [Binder](https://mybinder.org/v2/gh/pykale/pykale/digits-notebook?filepath=examples%2Fdigits_dann_lightn%2Fmain.ipynb)

bobturneruk:
Currently getting this error locally and on myBinder:

```
AttributeError: 'NoneType' object has no attribute 'best_model_path'
```

Related: https://github.com/PyTorchLightning/pytorch-lightning/issues/2547

Colab needs work so it clones the whole repo.

haipinglu:
> Currently getting this error locally and on myBinder:
> 
> ```
> AttributeError: 'NoneType' object has no attribute 'best_model_path'
> ```
> 
> Related: [PyTorchLightning/pytorch-lightning#2547](https://github.com/PyTorchLightning/pytorch-lightning/issues/2547)
> 
> Colab needs work so it clones the whole repo.

Did it work previously (when you said down to 5min)? 
If yes, it might be some config is too minimal that becomes out of expectation (e.g. possible due to min epoch=0?).

examples/digits_dann_lightn/main.ipynb
```diff
@@ -0,0 +1,191 @@
...
+    "    logger, results, checkpoint_callback, test_csv_file = setup_logger(\n",
+    "        train_params, cfg.OUTPUT.DIR, cfg.DAN.METHOD, seed\n",
+    "    )\n",
+    "\n",
+    "    if gpus is None:\n",
+    "        trainer = pl.Trainer(\n",
+    "            progress_bar_refresh_rate=cfg.OUTPUT.PB_FRESH,  # in steps\n",
+    "            min_epochs=cfg.SOLVER.MIN_EPOCHS,\n",
+    "            max_epochs=cfg.SOLVER.MAX_EPOCHS,\n",
+    "            checkpoint_callback=checkpoint_callback,\n",
```
sz144, line 143: Is your Pytorch Lightning>=1.3.0? Could you try replacing `checkpoint_callback=checkpoint_callback` with `callbacks=[checkpoint_callback],`. It probably can fix the above Attribute Error. Reference: https://pytorch-lightning.readthedocs.io/en/stable/common/weights_loading.html

bobturneruk: Thanks @sz144 - I'll check this out.

bobturneruk: This has helped and the notebook now runs, with Pytorch Lightning>=1.3.0 specified in `setup.py` for extras.

bobturneruk:
I suspect the tests are currently failing due to data download issues, but I'm not 100% sure.

bobturneruk:
Resolves https://github.com/pykale/pykale/issues/148

haipinglu:
> Resolves #148

I've linked it on the right.

examples/digits_dann_lightn/main.ipynb
sz144: @haipinglu we need check our previous examples for compatibility with the latest PyTorch Lightning.

bobturneruk: Or we could pin a specific earlier version.

haipinglu: Ideally, this should be covered by `tests` rather than manual check. Of course examples are not covered by tests but I think good tests should do this. The priority should be the API, then if possible we could cover the examples. But I do not think that this is the priority now.

If we found any compatibility issues, we can have respective tests to cover them or discuss how to prevent that from happening in future.

haipinglu:
@bobturneruk The TDC server recovered and all tests have passed now. 

Really like the progress bar now but the accuracy shows nothing is learned. Please consider the previous suggestion (using the emailed example) to take a subset of 2 classes for demo. In this case, we could use the saved 80%+ time for more epochs to have a better than random accuracy to show.

bobturneruk:
Thanks @haipinglu:

- Thanks for rerunning!
- I agree, if nothing is learned then it's perhaps merely "machine thinking"! I suggest that ideally this example is not the place to put code for splitting and reconstructing the example data into the right objects. Maybe we should cache the data we need in https://github.com/pykale/data? I'd welcome contributions from others on this task to speed things up.
- Yes - colab, by default, opens the notebook rather than building a container from the repository like binder.
- Yes. I'll have a think if there's a more elegant solution. It could be with the changes in this PR colab would need to install from this branch rather than pypi to run properly.

I'll prioritise cutting down the data and increase the epochs to get myBinder to perform better, unless I hear from you that you'd prefer me to look again at colab to deliver a speedup without altering the data, perhaps returning to myBinder afterwards.

haipinglu:
> *I suggest that ideally this example is not the place to put code for splitting and reconstructing the example data into the right objects. Maybe we should cache the data we need in https://github.com/pykale/data? I'd welcome contributions from others on this task to speed things up.

- I think the best solution is to follow the example notebook I sent to you via email to construct subset. It should take less than 10 lines of code. Have you tried the notebook? Constructing a new dataset will be much more involved (see mnistm.py and upsp.py)
- I do not think we are allowed to host data not owned by us (even if a subset) on pykale/data. We have to ask for permission/license unless the owner allows so explicitly.
- Subset extraction could be highly reusable for other examples and eventually become part of the core kale API.

> I'll prioritise cutting down the data and increase the epochs to get myBinder to perform better, unless I hear from you that you'd prefer me to look again at colab to deliver a speedup without altering the data, perhaps returning to myBinder afterwards.

Yes, let's get myBinder running first. For Colab, we can deal with it, maybe learn from some other successful examples to see how they did it. Thanks!

bobturneruk:
Another potential hosting option:

[View in Deepnote](https://deepnote.com/viewer/github/pykale/pykale/blob/digits-notebook/examples/digits_dann_lightn/main.ipynb)

bobturneruk:
Can't currently get it to work `Unable to duplicate the notebook.`.

haipinglu:
> Can't currently get it to work `Unable to duplicate the notebook.`.

I did not see it used by other ML/pytorch packages so I think we should stick to colab and/or binder, given the two weeks left till the deadline.

haipinglu:
@bobturneruk : @mustafa1728 has got the 2-class version done, pending integration. We can discuss later today.

bobturneruk:
Saw the notebook - looks good @mustafa1728!

bobturneruk:
I've learned a bit about pytorch from reading it!

mustafa1728:
> Saw the notebook - looks good @mustafa1728!

Thanks @bobturneruk. Glad to be of help!

bobturneruk:
Now runs (at least partially) on both Colab and myBinder without user needing to comment anything in or out. No progress bar in either (but works locally). Maybe related: https://github.com/PyTorchLightning/pytorch-lightning/issues/1112

bobturneruk:
Now fine and progress bar also OK.

bobturneruk:
I've added a fair bit of text to the notebook. A few ToDos in there, but could probably take some feedback on the work in progress.

bobturneruk:
Well done on https://github.com/pykale/pykale/pull/151!

haipinglu:
@bobturneruk #151 has been merged and class subset is ready for use. For example, check out https://github.com/pykale/pykale/blob/main/tests/loaddata/test_digits_access.py#L61

We need to have this merged this week for paper submission next week (which will use the notebook as an example). Many thanks!

bobturneruk:
Thanks @haipinglu. I will have some more time to work on this this week. If you have any feedback on the current form, or on https://github.com/pykale/pykale/pull/152, please share.

examples/digits_dann_lightn/configs/TUTORIAL.yaml
```diff
@@ -0,0 +1,14 @@
+DAN:
```
haipinglu, line 1: filename TUTORIAL, no need to be all uppercase. MN2UP-DANN etc are because they are abbreviations/acronyms of longer words.

examples/digits_dann_lightn/main.ipynb
```diff
@@ -0,0 +1,365 @@
+{
```
haipinglu, line 1: filename: how about calling it `tutorial.ipynb`. Two `mains` could be confusing to some.

examples/digits_dann_lightn/main.ipynb
```diff
@@ -0,0 +1,365 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a6e646cf",
+   "metadata": {},
+   "source": [
+    "## Train Model\n",
+    "\n",
+    "Optimise model parameters using the trainer."
```
haipinglu, line 276: Very minor: we agreed to follow US English

examples/digits_dann_lightn/main.ipynb
```diff
@@ -0,0 +1,365 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c311a3ac",
+   "metadata": {},
+   "source": [
+    "## Test Optimised Model\n",
+    "\n",
+    "Check performance of model optmised with training data against test data which was not used in training."
```
haipinglu, line 299: typo "optmised". Again a spell check plug-in may help.

examples/digits_dann_lightn/main.ipynb
```diff
@@ -0,0 +1,365 @@
...
+    ")\n",
+    "results.print_scores(cfg.DAN.METHOD)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "cdc6b6a1",
+   "metadata": {},
+   "source": [
+    "*ToDo: Do we need to say anything about how to interpret this output?*"
```
haipinglu, line 322: Yes, after updating the subset results.

haipinglu:
> Thanks @haipinglu. I will have some more time to work on this this week. If you have any feedback on the current form, or on #152, please share.

Thanks @bobturneruk. I have done some light comments and tested the new colab and binder. Colab works great and is much faster than binder. Colab started training already while binder is still spinning.

After updating the subset, hopefully we can get some useful results that are easy to explain. Then we can merge!

examples/digits_dann_lightn/configs/TUTORIAL.yaml
bobturneruk: Thanks, changes made.

examples/digits_dann_lightn/main.ipynb
bobturneruk: Good point. Changed.

examples/digits_dann_lightn/main.ipynb
```diff
@@ -0,0 +1,365 @@
...
+   "source": [
+    "model, train_params = get_model(cfg, dataset, num_channels)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7afdb064",
+   "metadata": {},
+   "source": [
+    "Output reposrts on data file use."
```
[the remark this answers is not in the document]
bobturneruk: Thanks.

examples/digits_dann_lightn/main.ipynb
bobturneruk: We did, thanks.

examples/digits_dann_lightn/main.ipynb
bobturneruk: Thanks. I now have one installed.

bobturneruk:
@haipinglu - I think this is now pretty close. Remaining issues from my perspective are:

- Not sure if I've described "Domain Adaptation" or our outputs correctly.
- Still pretty slow without GPU, but progress bars at least give the user some idea what's going on. Do we need to profile to see if subsetting actually speeds up?
- Some links will need to be changed to point to the `main` branch when this is merged.
- Is learning taking place (I think so)? Do we need more epochs?

haipinglu:
> * Not sure if I've described "Domain Adaptation" or our outputs correctly.

Add a link to wiki will be good. Then follow the top summary in wiki to describe: https://en.wikipedia.org/wiki/Domain_adaptation

> * Still pretty slow without GPU, but progress bars at least give the user some idea what's going on. Do we need to profile to see if subsetting actually speeds up?

Much more bearable with Colab. Yes, pbar is helpful. No need to profile now.

> * Some links will need to be changed to point to the `main` branch when this is merged.

OK.

> * Is learning taking place (I think so)? Do we need more epochs?

I saw 3-class got 44% >random 33%. I tried 2 classes just 55% marginally > random 50.  Tried 2 class with 0.1 split: still just 55.7%. We'd better tweak to a good acc.

+ Can we add colab and binder link (+icon) at the top of the notebook like many do?

Many thanks. We can target to do merging in today's meeting.

haipinglu:
@bobturneruk It will be useful (for us as well as users) to record and display the time taken for training, testing, and overall.

bobturneruk:
> Can we add colab and binder link (+icon) at the top of the notebook like many do?

Can you share an example, please, @haipinglu?

haipinglu:
> > Can we add colab and binder link (+icon) at the top of the notebook like many do?
> 
> Can you share an example, please, @haipinglu?

https://stackoverflow.com/questions/1557571/how-do-i-get-time-of-a-python-programs-execution

bobturneruk:
The button, not the timing. Timing probably a bit different in IPython. Speak soon!

haipinglu (approved):

PR #150, opened 2021-05-20
Test gripnet

pz-white:
### Description
- Add gripnet test case
- Fix check_transform issue raised by rdkit version
- Update gripnet example and now the data can be automatically get from pykale/data instead downloading mannually

### Status
**Ready**

### Types of changes


pz-white:
The GripNet test case is compact as I re-wrote the data construction and now the processed data has been put on pykale/data.

tests/prepdata/test_chem_transform.py
```diff
@@ -3,7 +3,7 @@
 
 def test_chem_wrong_smiles():
```
haipinglu, line 4: Are you sure that no assertion is needed for this test?

haipinglu (approved):

PR #151, opened 2021-06-04
Add Class subsampling

mustafa1728:
Associated with https://github.com/pykale/pykale/pull/149#issuecomment-851938657.

### Description
Adding class sub-sampling to core kale API. 
Class sub-sampling will allow datasets to easily sample only a subset of classes. 

### API update description
Adding `class_ids=[id_0, id_1 ...]` parameter in initialisations of: 
- MultiDomainDatasets (as being used in [digits_dann_lightn](https://github.com/pykale/pykale/tree/main/examples/digits_dann_lightn) example) 
- VideoMultiDomainDatasets (as being used in [action_dann_lightn](https://github.com/pykale/pykale/tree/main/examples/action_dann_lightn) example) 

should make it use class sub-sampled data. With no parameter or None value, dataset will use all classes.

### Status
**Ready**

### Types of changes


kale/loaddata/dataset_access.py
```diff
@@ -28,21 +28,49 @@ def get_train(self):
         """
         raise NotImplementedError()
 
-    def get_train_val(self, val_ratio):
+    def get_class_subsampled_train(self, class_ids=None):
```
haipinglu, line 31: Possible to have a neater name? Also possible to minimize changes in naming? 
Suggestions to consider: `get_train_class_subset`, e.g. in a [CVPRW19 paper](https://openaccess.thecvf.com/content_CVPRW_2019/papers/Weakly%20Supervised%20Learning%20for%20Real-World%20Computer%20Vision%20Applications/Zohrizadeh_Class_Subset_Selection_for_Partial_Domain_Adaptation_CVPRW_2019_paper.pdf)

kale/loaddata/dataset_access.py
```diff
@@ -28,21 +28,49 @@ def get_train(self):
...
+        """
+        Args:
+            class_ids (list, optional): List of class ids that are to be subsampled.
```
haipinglu, line 34: Here we are more choosing a subset (purposely) rather than doing subsampling. Consider to use subset, e.g. class_ids are List of the chosen subset of class ids

kale/loaddata/dataset_access.py
```diff
@@ -28,21 +28,49 @@ def get_train(self):
...
+        Returns: a torch.utils.data.Dataset
+            Dataset: a torch.utils.data.Dataset with only classes in class_ids
+        """
+        train_dataset = self.get_train()
+        if class_ids is None:
+            return train_dataset
+        else:
+            sub_indices = [i for i in range(0, len(train_dataset)) if train_dataset[i][1] in class_ids]
+            return torch.utils.data.Subset(train_dataset, sub_indices)
```
haipinglu, line 43: See here torch also use subset rather than subsample, try to follow torch naming conventions as much as possible.

kale/loaddata/dataset_access.py
```diff
@@ -28,21 +28,49 @@ def get_train(self):
...
+
+    def get_train_val(self, val_ratio, sub_class_ids=None):
```
haipinglu, line 45: Inconsistency. class_ids above is now called sub_class_ids. Be consistent (the shorter version should be fine).

kale/loaddata/dataset_access.py
```diff
@@ -28,21 +28,49 @@ def get_train(self):
...
         """
         Randomly split a dataset into non-overlapping training and validation datasets.
 
         Args:
             val_ratio (float): the ratio for validation set
-
+            sub_class_ids (list, optional): List of class ids that are to be subsampled.
```
haipinglu, line 51: See above

kale/loaddata/dataset_access.py
```diff
@@ -28,21 +28,49 @@ def get_train(self):
...
+        train_dataset = self.get_class_subsampled_train(sub_class_ids)
         ntotal = len(train_dataset)
         ntrain = int((1 - val_ratio) * ntotal)
 
         return torch.utils.data.random_split(train_dataset, [ntrain, ntotal - ntrain])
 
     def get_test(self):
         raise NotImplementedError()
+
+    def get_class_subsampled_test(self, class_ids):
```
haipinglu, line 64: See comments above for train

kale/loaddata/dataset_access.py
```diff
@@ -28,21 +28,49 @@ def get_train(self):
...
+            class_ids (list, optional): List of class ids that are to be subsampled.
+        Returns: a torch.utils.data.Dataset
+            Dataset: a torch.utils.data.Dataset with only classes in class_ids
+        """
+        test_dataset = self.get_test()
+        if class_ids is None:
+            return test_dataset
+        else:
+            sub_indices = [i for i in range(0, len(test_dataset)) if test_dataset[i][1] in class_ids]
+            return torch.utils.data.Subset(test_dataset, sub_indices)
```
haipinglu, line 76: Line 72-76 is somehow a repetition of line 39-43. Is it possible and worthy to reduce such repetition (e.g. define that 4 lines as a function taking ids and dataset in)?

kale/loaddata/dataset_access.py
haipinglu: Not subclass (meaning finer classification) is different from a subset of classes.

kale/loaddata/dataset_access.py
haipinglu: Also note subclass (meaning finer classification) is different from a subset of classes.

kale/loaddata/multi_domain.py
```diff
@@ -68,6 +68,7 @@ def __init__(
         target_sampling_config=None,
         n_fewshot=None,
         random_state=None,
+        sub_class_ids=None,
```
haipinglu, line 71: See about for all. sub_class could be misleading. Subclasses are finer classes of a class.

kale/loaddata/video_access.py
```diff
@@ -224,11 +224,11 @@ def __init__(
         self._transform = video_transform.get_transform(transform_kind, self._image_modality)
         self._seed = seed
 
-    def get_train_val(self, val_ratio):
+    def get_train_val(self, val_ratio, sub_class_ids=None):
```
haipinglu, line 227: See above

kale/loaddata/video_multi_domain.py
```diff
@@ -29,6 +29,7 @@ def __init__(
         target_sampling_config=None,
         n_fewshot=None,
         random_state=None,
+        sub_class_ids=None,
```
haipinglu, line 32: Likewise

tests/loaddata/test_digits_access.py
```diff
@@ -15,6 +15,8 @@
 DATASIZE_TYPE = ["max", "source"]
 VAL_RATIO = [0.1]
 
+CLASS_SUB_SAMPLES = [[1, 3, 8]]
```
haipinglu, line 18: See above: subset not samples

tests/loaddata/test_digits_access.py
```diff
@@ -51,3 +53,38 @@ def test_get_train_test(dataset_name, download_path):
...
+    dataset_subsampled = MultiDomainDatasets(
+        source,
+        target,
+        config_weight_type=WEIGHT_TYPE[0],
+        config_size_type=DATASIZE_TYPE[1],
+        sub_class_ids=class_sub_sample,
+    )
+    dataset_subsampled.prepare_data_loaders()
+
+    assert len(dataset_subsampled) <= len(dataset)
```
haipinglu, line 81: You can have stronger assertions here. Since these digit datasets have the same number of samples for each class, we should have len(dataset_subsampled) == 0.3*len(dataset) since you take 3 out of 10 classes, right?

tests/loaddata/test_digits_access.py
```diff
@@ -51,3 +53,38 @@ def test_get_train_test(dataset_name, download_path):
...
+    assert len(source_train) <= len(source.get_train())
+    assert len(source_test) <= len(source.get_test())
+
+    assert isinstance(source_train, torch.utils.data.Dataset)
```
haipinglu, line 85: These assertions are less useful, right? We used similar assertions when we do not have stronger assertions implementation. In your case, the above can be stronger assertions. Consider to remove these.

tests/loaddata/test_digits_access.py
```diff
@@ -51,3 +53,38 @@ def test_get_train_test(dataset_name, download_path):
     assert source.n_classes() == 10
     assert isinstance(source_train, torch.utils.data.Dataset)
     assert isinstance(source_test, torch.utils.data.Dataset)
+
+
+@pytest.mark.parametrize("dataset_name", ALL)
```
haipinglu, line 58: It may not be necessary to loop through all variations, if they were tested elsewhere.  We can talk tomorrow.

tests/loaddata/test_video_access.py
```diff
@@ -115,3 +120,37 @@ def test_get_source_target(source_cfg, target_cfg, val_ratio, weight_type, datas
...
+        if dataset.flow:
+            dataset._flow_source_by_split = {"train": train_val[0]}
+            dataset._flow_target_by_split = {"train": train_val[0]}
+        if dataset_subsampled.rgb:
+            dataset_subsampled._rgb_source_by_split = {"train": subsampled_train_val[0]}
+            dataset_subsampled._rgb_target_by_split = {"train": subsampled_train_val[0]}
+        if dataset_subsampled.flow:
+            dataset_subsampled._flow_source_by_split = {"train": subsampled_train_val[0]}
+            dataset_subsampled._flow_target_by_split = {"train": subsampled_train_val[0]}
+        assert len(dataset_subsampled) <= len(dataset)
```
haipinglu, line 156: See the comments for digits. Stronger versions are possible.

tests/loaddata/test_video_access.py
```diff
@@ -115,3 +120,37 @@ def test_get_source_target(source_cfg, target_cfg, val_ratio, weight_type, datas
...
+
+    # test class sub-sampling
+    if source_cfg == SOURCES[1] and target_cfg == TARGETS[1]:
+        subsampled_train_val = source["rgb"].get_train_val(val_ratio, class_sub_sample)
+        assert isinstance(subsampled_train_val, list)
+        assert isinstance(subsampled_train_val[0], torch.utils.data.Dataset)
+        assert isinstance(subsampled_train_val[1], torch.utils.data.Dataset)
+
+        assert len(subsampled_train_val[0]) <= len(train_val[0])
+        assert len(subsampled_train_val[1]) <= len(train_val[1])
```
haipinglu, line 132: See the comments for digits. Stronger versions are possible.

kale/loaddata/dataset_access.py
mustafa1728: Ohh right, I did not know about subclasses. Sure l will update this. Thanks for pointing it out!

haipinglu (commented):
Good job. Please see the comments. If any further clarification is needed, we can discuss tomorrow (Monday) and/or Tuesday. 
You can mark it as `Ready for review` (bottom of PR) now.

@XianyuanLiu Please take a look too before we meet on Monday.

kale/loaddata/dataset_access.py
mustafa1728: Yes I feel that is definitely worthy. I am thinking a private _get_subset function? Also, I wanted to ask whether this would be alright since this function will come in the 'read the docs' as well.

kale/loaddata/dataset_access.py
```diff
@@ -28,21 +28,53 @@ def get_train(self):
...
 
         Args:
             val_ratio (float): the ratio for validation set
-
+            class_ids (list, optional): List of chosen subset of class ids.
         Returns:
             Dataset: a torch.utils.data.Dataset
         """
-        train_dataset = self.get_train()
+        train_dataset = self.get_train_class_subset(class_ids)
```
XianyuanLiu, line 50: We may give a flag `class_ids` to `get_train()` directly without changing `self.get_train()` here.

kale/loaddata/dataset_access.py
```diff
@@ -28,21 +28,53 @@ def get_train(self):
...
+    def get_test_class_subset(self, class_ids):
+        """
+        Args:
+            class_ids (list, optional): List of chosen subset of class ids.
+        Returns: a torch.utils.data.Dataset
+            Dataset: a torch.utils.data.Dataset with only classes in class_ids
+        """
+        return self._get_subset(self.get_test(), class_ids)
+
+    def _get_subset(self, dataset, class_ids):
```
XianyuanLiu, line 68: In my opinion, `_get_subset` for class can be a simple and direct function to get the subset without conditionals. We can put the conditionals in `get_train` and `get_test`. It will be more clear to understand.

tests/loaddata/test_digits_access.py
XianyuanLiu: Agree with Haiping. In addition, my most assertions are too general and simple and only check some parameter type, because we want to achieve high coverage in the first step. However, from now on, we are going to improve our test example by replacing some assertions with stronger ones. Therefore, in your test example, you can explore assertions for a more detailed test than the current code.

XianyuanLiu (changes requested):
Well done. Need to reorganize the code. If you have other ideas, please comment.

kale/loaddata/multi_domain.py
```diff
@@ -123,25 +125,26 @@ def __init__(
...
 
         logging.debug("Load target")
         (self._target_by_split["train"], self._target_by_split["valid"],) = self._target_access.get_train_val(
-            self._val_split_ratio
+            self._val_split_ratio, self.class_ids
         )
 
         logging.debug("Load source Test")
-        self._source_by_split["test"] = self._source_access.get_test()
+        self._source_by_split["test"] = self._source_access.get_test_class_subset(self.class_ids)
```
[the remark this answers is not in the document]
mustafa1728: I think the issue with this is that `get_train()` and `get_test()` are being redefined by its child classes. So, similar flags would have to be handled by each of the child classes separately. `get_train_val()` is not being re-defined, so I have added flags inside it.

kale/loaddata/dataset_access.py
mustafa1728: Yes agreed. I will update this, thanks.

kale/loaddata/dataset_access.py
haipinglu: Code will be reviewed and can be updated so you may make the changes good in your opinion and we will review. When not sure, check how pytorch implements something similar to learn.

kale/loaddata/multi_domain.py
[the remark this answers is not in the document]
mustafa1728: I think another option could be to have the child classes redefine a different function (maybe `_get_train()` private) and call this inside `get_train()`?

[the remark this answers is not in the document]
XianyuanLiu: > I think another option could be to have the child classes redefine a different function (maybe `_get_train()` private or `get_train_all()`) and call this inside `get_train()` with class_id flags?

It will be a good idea if feasible.

mustafa1728:
A simpler solution could be to keep a single separate function for getting class-subsets and using this function in `MultiDomainDatasets` and `VideoMultiDomainDatasets`. The advantage is that the base `DatasetAccess` class is not modified and it becomes a bit easier to understand as well. The last two commits are in this direction and this separate function is in `kale.utils.class_subset`.

kale/loaddata/video_multi_domain.py
```diff
@@ -91,29 +92,67 @@ def prepare_data_loaders(self):
...
+                    self._flow_target_by_split["valid"], self.class_ids
+                )
 
             logging.debug("Load flow Test")
-            self._flow_source_by_split["test"] = self._source_access_dict["flow"].get_test(self.class_ids)
-            self._flow_target_by_split["test"] = self._target_access_dict["flow"].get_test(self.class_ids)
+            self._flow_source_by_split["test"] = self._source_access_dict["flow"].get_test()
+            self._flow_target_by_split["test"] = self._target_access_dict["flow"].get_test()
+            if self.class_ids is not None:
+                self._flow_source_by_split["test"] = get_class_subset(
```
XianyuanLiu, line 150: Could we put this function in the `DatasetAccess`? Is it feasible?
For example,
self._flow_source_by_split["test"] = self._flow_source_by_split["test"].get_class_subset(self.class_ids)

XianyuanLiu (changes requested):
Good job!

It will be better to put `get_class subset` in the class `Dataset Access` if possible. I think it will be easier for people to find it. `get_class subset` is also an enhancement for `Dataset Access`.

kale/loaddata/video_multi_domain.py
mustafa1728: Yes agreed, that will be better. Thanks, I'll update it.

mustafa1728: Ohh actually it may not be feasible. So, `self._flow_source_by_split["test"]` is actually a normal `torch.Dataset` and not a DatasetAccess object. 

If this function was to be added in the class, it could be a static function, and used something like: `DatasetAccess. get_class_subset(dataset, class_ids)`

mustafa1728: Ohh, so it may not be feasible as is, since `self._flow_source_by_split["test"]` is a simple` torch.Dataset` object and not a `DatasetAccess` object.

The function can be put inside DatasetAccess class as a static function, so it will be used like: `dataset = DatasetAccess.get_class_subset(dataset, class_ids)`. Would this be appropriate?

XianyuanLiu: Oh, I ignore the type. No need to change. That seems a little weird :) The current one is okay but it only has a few lines in `kale/utils/class_subset.py`. Do we have any other choice to put this function? We can keep the current now and discuss with Haiping at the meeting. Thanks!

XianyuanLiu (approved):
Well done! Thanks!

tests/loaddata/test_digits_access.py
```diff
@@ -51,3 +54,33 @@ def test_get_train_test(dataset_name, download_path):
     assert source.n_classes() == 10
     assert isinstance(source_train, torch.utils.data.Dataset)
     assert isinstance(source_test, torch.utils.data.Dataset)
+
+
+@pytest.mark.parametrize("class_subset", CLASS_SUBSETS)
+@pytest.mark.parametrize("val_ratio", VAL_RATIO)
+def test_class_subsampling(class_subset, val_ratio, download_path):
```
haipinglu, line 61: try to avoid using "subsampling"

XianyuanLiu (approved):

PR #155, opened 2021-06-12
Update docs for next release & restore MNIST

haipinglu:
### Description
- Update documentations (including fixing typos etc).
- Restore MNIST tests

### Status
**Ready**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).

PR #156, opened 2021-06-14
Empty tests and remove MNIST test

bobturneruk:
### Description
Remove empty test files and remove MNIST test.

### Status
**Ready**

### Types of changes


bobturneruk:
I think we're hitting MNIST download errors again which are causing the tests to fail. They pass locally (Windows 10).

haipinglu:
> I think we're hitting MNIST download errors again which are causing the tests to fail. They pass locally (Windows 10).

I am aware of it. I added MNIST tests back in #155. It was not as bad as before. We can remove it if we cannot bear it. In merging 155, it was bearable. I reran 2 or 3 times to pass.

haipinglu (approved):
I removed MNIST test. It does fails more often than SVHN and USPS. Losing 1% coverage is not too bad.

Issue #182, opened 2021-07-13
Google Drive download API


haipinglu:
@RaivoKoot Can you create a Google Drive download API from your video loading example like what Xianyuan did for regular file download? Thanks.

RaivoKoot:
Should be possible. Will try. Am very swamped with internship and more work right now. So, might take a little bit longer than usual. Will do it though.

RaivoKoot:
ETA: 1 week from today

PR #196, opened 2021-08-01
Add Google Drive Download API

mustafa1728:
Fixes https://github.com/pykale/pykale/issues/182.

### Description
Adds drive download function from [video loading example](https://github.com/pykale/pykale/blob/main/examples/video_loading/main.py) to core API for better reusability. API allows easier download of google drive files from their ids.

### Status
**Ready**

### Types of changes

### DIFF
--- a/tests/utils/test_download.py
+++ b/tests/utils/test_download.py
@@ -13,6 +13,7 @@ PARAM = [
 
 GDRIVE_PARAM = [
     "1U4D23R8u8MJX9KVKb92bZZX-tbpKWtga;demo_datasets.zip;zip",
+    "1SV7fmAnWj-6AU9X5BGOrvGMoh2Gu9Nih;dummy_data.csv;csv",
 ]
 
 
--- a/tests/utils/test_download.py
+++ b/tests/utils/test_download.py
@@ -3,12 +3,16 @@ from pathlib import Path
 
 import pytest
 
-from kale.utils.download import download_file_by_url
+from kale.utils.download import download_file_by_url, download_file_gdrive
 
 output_directory = Path().absolute().parent.joinpath("test_data/download")
 PARAM = [
     " https://github.com/pykale/data/raw/main/video_data/video_test_data/ADL/annotations/labels_train_test/adl_P_11_train.pkl;a.pkl;pkl",
     " https://github.com/pykale/data/raw/main/video_data/video_test_data.zip;video_test_data.zip;zip",
+]
+
+GDRIVE_PARAM = [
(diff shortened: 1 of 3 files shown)


XianyuanLiu:
This example is grabbing a zip file from google drive. Could you add another example with no need to uncompress like URL one?

XianyuanLiu (approved):
Well done. LGTM. Thanks.

tests/utils/test_download.py
```diff
@@ -3,14 +3,19 @@
...
 
 output_directory = Path().absolute().parent.joinpath("test_data/download")
 PARAM = [
     " https://github.com/pykale/data/raw/main/video_data/video_test_data/ADL/annotations/labels_train_test/adl_P_11_train.pkl;a.pkl;pkl",
     " https://github.com/pykale/data/raw/main/video_data/video_test_data.zip;video_test_data.zip;zip",
 ]
 
+GDRIVE_PARAM = [
+    "1U4D23R8u8MJX9KVKb92bZZX-tbpKWtga;demo_datasets.zip;zip",
+    "1SV7fmAnWj-6AU9X5BGOrvGMoh2Gu9Nih;dummy_data.csv;csv",
```
haipinglu, line 16: Where are these located? Which Google account? Permanent?

haipinglu (changes requested):
Thanks. The video loading example should be updated to use this API and remove duplication. See the other question.

tests/utils/test_download.py
mustafa1728: The first one was taken directly from the video loading example, and is in Raivo's Google account. The second one is in my account.

These are not permanent. Is there a PyKale google account? Maybe these dummy files can be added there.

mustafa1728:
> The video loading example should be updated to use this API and remove duplication. 

So may I edit that example in this PR itself or would Raivo be doing that?

haipinglu:
> > The video loading example should be updated to use this API and remove duplication.
> 
> So may I edit that example in this PR itself or would Raivo be doing that?

haipinglu:
> > The video loading example should be updated to use this API and remove duplication.
> 
> So may I edit that example in this PR itself or would Raivo be doing that?

Please do it @mustafa1728. Thanks.

tests/utils/test_download.py
haipinglu: Thanks for the clarification. It is fine then. We do not have a PyKale Google account (yet). No need to create just for dummy files.

haipinglu (approved):

PR #228, opened 2021-10-08
Zenodo json

bobturneruk:
### Description
Add json metadata to enrich "publication" of releases via Zenodo (which assigns a DOI and makes the software more compliant with FAIR priniciples).

### Types of changes

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -35,6 +35,7 @@
             "name": "Shuo Zhou"
         },
         {
+            "orcid": "0000-0001-8694-3453",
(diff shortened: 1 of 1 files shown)


bobturneruk:
Not sure if the file I've added is yet schema compliant. I will need to get details (e.g. orcids) from other authors before this leaves draft status. 

Zenodo is currently configured to automatically pull what metadata it can from this repository to make a DOI identifiable record on each release. With this PR merged, it will use the richer metadata set from `.zenodo.json`.

PR #282, opened 2021-12-22
Clarify when to request review and prefer just one label

haipinglu:
### Description
Clarify when to request review and prefer only one label for easy categorization of changes.

### Status
**Ready**

### Types of changes


haipinglu:
@Schobs See if this version is clearer that review should be requested only when the PR is ready.

Schobs (approved):
Looks good. It is more clear now.

PR #302, opened 2022-03-15
Torch cluster bug

pz-white:
Fixes #299 

### Description
Fix issues #299 

### Status
**Work in progress**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).

PR #298, opened 2022-02-10
Retry all tests

bobturneruk:
Fixes #278

### Description
Current approach is to retry the tests as part of the workflow prior to formally failing.

### Status
**Work in progress**

If you've been given access to pykale with role Triage or above, on the right (delete these after selection):

- **When** your pull request is **ready** for review, select a reviewer. Use the suggested one if unsure.

### Types of changes
- [x] Non-breaking change (fix or new feature that would not break existing functionality).

PR #284, opened 2022-01-13
Update DICOM reading and image visualization

sz144:
### Description

- Add two dependencies python-gdcm and pylibjpeg for fix the issue of reading lossless JPEG compressed in DICOM files.
- Allow reading dicom files with string in PatientID.
- Split dicom reading as three modular functions.
- Allow specifying colours and adding annotation for markers in image visualization.

### Status
**Ready**

### Types of changes

- [ ] Non-breaking change (fix or new feature that would not break existing functionality).
- [x] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [x] In-line docstrings updated.
- [ ] [Source for documentation at `docs`](https://github.com/pykale/pykale/tree/main/docs/source) manually updated for new API.

PR #304, opened 2022-03-16
Re target tests

bobturneruk:
Fixes *NA*.

### Description
Moves CI tests to a better range of python versions.

### Status
**Completed**

If you've been given access to pykale with role Triage or above, on the right (delete these after selection):

- **When** your pull request is **ready** for review, select a reviewer. Use the suggested one if unsure.

If you cannot see an option to select a reviewer/label, that means one maintainer will get notified upon your pull request and then select for you.

### Types of changes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ setup(
         "Source": "https://github.com/pykale/pykale",
     },
     packages=find_packages(exclude=("docs", "examples", "tests")),
-    python_requires=">=3.7, <=3.9",
+    python_requires=">=3.7, <=3.8",
     install_requires=requirements,
     extras_require={"dev": dev_requirements},
     setup_requires=["setuptools>=38.6.0"],
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ setup(
         "Source": "https://github.com/pykale/pykale",
     },
     packages=find_packages(exclude=("docs", "examples", "tests")),
-    python_requires=">=3.6",
+    python_requires=">=3.7, <=3.9",
     install_requires=requirements,
     extras_require={"dev": dev_requirements},
     setup_requires=["setuptools>=38.6.0"],
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,4 +64,4 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           filename: .github/ISSUE_TEMPLATE/action-fail.md
-        if: ${{ always() && steps.run_tests.outcome == 'failure' && github.event_name == 'schedule'}}
+        if: ${{ always() && contains('failure', steps.*.outcome)  && github.event_name == 'schedule'}}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,14 +19,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9"]
         exclude:
           - os: windows-latest
             python-version: "3.7"
           - os: windows-latest
             python-version: "3.8"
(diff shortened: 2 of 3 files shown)


bobturneruk:
@haipinglu - seems that conda forge rdkit doesn't support python 3.10 yet. Shall we stick with 3.9 as lastest supported python for pykale for now?

https://github.com/pykale/pykale/runs/5568498604?check_suite_focus=true
https://github.com/pykale/pykale/runs/5568498604?check_suite_focus=true

bobturneruk:
Problems on Windows with Python 3.9 (related https://github.com/pykale/pykale/pull/302). cc @pz-white

haipinglu:
@bobturneruk For the purpose of testing, it is fine if higher versions of python are not yet supported. We expect these issues to be resolved later by the related libraries. I think it is more important to fix what we can fix, particularly #298 to solve issues like #305. Many thanks.

haipinglu:
> Problems on Windows with Python 3.9 (related #302). cc @pz-white

Hi @bobturneruk, you may notice that Peizhen has fixed the problem. Can you give an update please?

bobturneruk:
@haipinglu - it's failing on windows + python 3.9 here. I'll try a rebase.

haipinglu:
> @haipinglu - it's failing on windows + python 3.9 here. I'll try a rebase.

@bobturneruk Sorry for not understanding the problem completely. The other PR on test failure should be more urgent as it regularly creates unnecessary issues. Could you fix that one first? Many thanks.
@pz-white Can you take a look whether the problem is fixable?

pz-white:
@haipinglu @bobturneruk 
As the pytorch we used is 1.7.0, the highest supported python is 3.8 for torch-cluster in this dependence. I think the problem is not fixable in the current stage, thanks.

bobturneruk:
Thanks @pz-white! I'll restrict the tests to 3.7 and 3.8, and make this limitation clear in docs etc.

bobturneruk:
Going to mark this as ready because I expect the tests will pass.

bobturneruk:
The Python 3.8 CI runs seem to be hanging on installing dependencies:

https://github.com/pykale/pykale/runs/5691738676?check_suite_focus=true

Any thoughts @pz-white? Perhaps we need to stick with 3.7 only?

haipinglu:
@bobturneruk I "reactivated" tests for python 3.7 and 3.8. This PR could be the one to fix the python version issue. Or, shall we open a new PR for fixing the version issue?

bobturneruk:
@haipinglu I've amended the text to show that all we are doing here is targetting the tests to versions of python that are supported and we are compatible with.

setup.py
```diff
@@ -88,7 +88,7 @@ def find_version(*file_paths):
         "Source": "https://github.com/pykale/pykale",
     },
     packages=find_packages(exclude=("docs", "examples", "tests")),
-    python_requires=">=3.6",
+    python_requires=">=3.7, <=3.8",
```
bobturneruk, line 91: As it stands this will prevent installation on any python other than 3.7 or 3.8. @haipinglu should we keep 3.6 here and not in the tests?

bobturneruk:
I suggest this PR is closed without merging and an issue raised to add compatibility for later python versions. Does this sound right, @haipinglu?

setup.py
haipinglu: So far, only 3.6 can pass the tests so we can keep 3.6 until we make 3.7, 3.8, or 3.9 work. You may also think whether we should have a fresh new PR or continue in this PR.

haipinglu:
> I suggest this PR is closed without merging and an issue raised to add compatibility for later python versions. Does this sound right, @haipinglu?

Yes, please proceed. Thanks.

haipinglu:
@bobturneruk I switched the required tests on 3.7 and 3.8 off. Otherwise, no update can be merged until we fix those issues, which we are not sure how long it will take yet (e.g. Shuo's PR #284).

bobturneruk:
Superseded by https://github.com/pykale/pykale/issues/316

PR #345, opened 2022-08-25
Add file path as a dicom dataset attribution

sz144:
### Description
Add file path as a dicom dataset attribution.

### Status
**Ready**

### Types of changes


Schobs (approved):
Great change. Now when you load a DICOM, each phase image is saved along with an attribute called "FilePath" which links the filepath to the loaded image.

PR #347, opened 2022-09-09
Fix rdkit error in testflow

pz-white:
### Description
Fix rdkit error in testflow

### Status
**Work in progress**

### Types of changes


pz-white:
Fix the rdkit test error raised by the new version of PyTDC=0.3.7. I downgraded to the original 0.3.6 since the new version has a new dependency of rdkit-pypi, which raised a potential compatibility problem with Windows OS (above python 3.7).

haipinglu (approved):

PR #354, opened 2022-10-19
Fix conda rdkit and boost-cpp conflict issue

pz-white:
### Description
Fix conda rdkit and boost-cpp conflict issue

### Status
**Work in progress**

If you've been given access to pykale with role Triage or above, on the right (delete these after selection):

### Types of changes

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-          pip install rdkit
+          conda install -c conda-forge rdkit=2022.03.5
           pip install pyparsing==2.4.7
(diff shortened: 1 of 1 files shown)


pz-white:
The error might be caused by the build of boost-cpp=1.74.0. I found a very similar issue from conda-forge/rdkit-feedstock: https://github.com/conda-forge/rdkit-feedstock/issues/93.

The strange thing is that we also test with boost-cpp 1.74.0 in previous successful build. Not sure if conda-forge happens the issue again now.

pz-white:
The same issue is reported on the conda-forge/rdkit-feedstock: https://github.com/conda-forge/rdkit-feedstock/issues/111

haipinglu:
The problem is gone after about 5 days, likely being fixed elsewhere in the ecosystem. This is a lesson for similar problems in future.

PR #365, opened 2023-04-18
Update project.yml

Mdnaimulislam:
### Description
Updated the project.yml action file for the new project board.

### Status
**Ready**

### Types of changes


haipinglu:
@Mdnaimulislam Creation of new label needs to be discussed and agreed before creation. `minor` is not a good label as other labels may be minor too so I have deleted that label.

.github/workflows/test_project.yml
```diff
@@ -0,0 +1,76 @@
+name: Add issues and PRs to project and move issues on label change
```
haipinglu, line 1: File name `test_project` is inconsistent with those in https://github.com/pykale/pykale/tree/main/.github/workflows where dash is used rather than underscore. Moreover, the name is questionable. I think it should replace https://github.com/pykale/pykale/blob/main/.github/workflows/project.yml. Therefore, it maybe better to update that yml rather than creating a new one. I understand that you may want to test it but I think we can merge and test since this is not a stable released version.

haipinglu: The formatting is also inconsistent with other workflow yml files.

.github/workflows/test_project.yml
```diff
@@ -0,0 +1,76 @@
...
+
+  move-issue-to-in-progress:
+    name: Move issue to In Progress column when work-in-progress label is added
+    if: github.event_name == 'issues' && github.event.action == 'labeled' && contains(github.event.issue.labels.*.name, 'work-in-progress')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Move issue to In Progress column
+        uses: actions/github-script@v5
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
haipinglu, line 43: Duplicate necessary? See `env:` in https://github.com/pykale/pykale/blob/main/.github/workflows/project.yml

haipinglu (changes requested):
Please update https://github.com/pykale/pykale/blob/main/.github/workflows/project.yml instead of creating a new one. Thanks a lot!

.github/workflows/project.yml
```diff
@@ -1,24 +1,80 @@
...
-# https://github.com/marketplace/actions/assign-to-one-project
+# This workflow will automatically assign issues and pull requests to the new default project and relocate issues to the project based on changes in their labels.
+# https://github.com/actions/add-to-project
 
 name: assign-project
 
 on:
   issues:
-    types: [opened, labeled]
+    types:
```
haipinglu, line 8: Is this change necessary? The same as below. We can talk tomorrow.

.github/workflows/project.yml
```diff
@@ -1,24 +1,80 @@
...
+      - labeled
+
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
-  assign-one-project:
-    name: Assign to one default project
+  add-issue-to-project:
+    name: Assign NEW issue to project with To do status
```
haipinglu, line 21: Better to enclose those column names with "" so it is clearer that they are column names, as you did at the bottom. For all.

.github/workflows/project.yml
```diff
@@ -1,24 +1,80 @@
...
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@v0.5.0
+        with:
+          project-url: https://github.com/orgs/pykale/projects/4
+          column-name: In progress
+
+  move-issue-to-in-progress:
+    name: Move issue to In progress column when work-in-progress label is assigned
```
haipinglu, line 41: Same here. Enclose label name with "". Check all please.

.github/workflows/project.yml
```diff
@@ -1,24 +1,80 @@
...
-      uses: srggrs/assign-one-project-github-action@1.2.1
-      if: github.event.action == 'opened'
-      with:
-        project: "https://github.com/pykale/pykale/projects/10"
+      - name: Move issue to In progress column
+        uses: actions/github-script@v5
+        with:
+          script: |
+            const { repo, payload } = context;
+            const projectId = 4;
```
haipinglu, line 50: At the top readme, specify rows that need change when we update the project board number.

.github/workflows/project.yml
```diff
@@ -1,24 +1,80 @@
...
+            });
+
+            // Find the "In progress" column
+            const inProgressColumn = columns.data.find(column => column.name === columnName);
+
+            if (!inProgressColumn) {
+              throw new Error(`Column "${columnName}" not found in project "${project.data.name}"`);
+            }
+
+            // Move the issue to the "In Progress" column
```
haipinglu, line 75: "In Progress" or "In progress"?
Check capitalisation and be consistent, for all.

haipinglu (changes requested):
More style issues to fix please. Thanks.

haipinglu (approved):

PR #436, opened 2024-05-27
Fix bugs of greek letters in docstring.

wenruifan:
### Description
Fix the bug in the docstring caused by the Greek letter in maths ``` :math:`\beta` ``` in `kale.pipeline.domain_adapter.WDGRLTrainer` and `kale.pipeline.domain_adapter.WDGRLTrainerMod`. It is changed to ``` :math:`\\beta` ``` (`\\` instead of `\` ). The documentation of `kale.pipeline.domain_adapter` can work now.

There will be an error when Greek letters start with escape sequences in strings like `\b`, `\a`, `\v`, `\t`, `\n`, etc. For example, using ``` :math:`\beta` ``` leads to `Maths input error` in the documentation, and using ``` :math:`\beta` ``` and ``` :math:`\gamma` ``` together leads to `Error: \g is not an escape sequence` in the `pre-commit` test.

### Note

The docstring's pattern of compiling `\\` and `\` in maths, especially Greek letters, is messy. The same code can work in some places but may fail elsewhere. I tried all Greek letters under multiple circumstances but failed to find a consistent pattern. Other commands like `\cdot` may also lead to some errors.

For example, ``` :math:`\\sigma` ``` can work in my experiments but cannot work in `kale.embed.mogonet.MogonetGCNConv()`. Instead, ``` :math:`\sigma` ``` can work in `kale.embed.mogonet.MogonetGCNConv()`.

Therefore, **please proofread the documentation after updating it**. 

If contributors encounter errors when they use maths, switching between `\\` and `\` could be a solution.

### Status
**Ready**

### Types of changes


haipinglu (changes requested):
@wenruifan Thanks. Add a **brief** version of this note to https://github.com/pykale/pykale/blob/main/.github/CONTRIBUTING.md#coding-style under the docstring bullets so that this is captured in the guideline rather than embedded in a (merged) PR. Talk to me tomorrow if unsure.

PR #437, opened 2024-07-05
Add mida_trainer pipeline

shuo-zhou:
### Description
Add a scikit-learn compatible transformer->mida->estimator pipeline 

### Status
**Ready**

### Types of changes

- [x] [Source for documentation at `docs`](https://github.com/pykale/pykale/tree/main/docs/source) manually updated for new API.


tests/pipeline/test_mida_trainer.py
```diff
@@ -0,0 +1,60 @@
...
+from sklearn.linear_model import LogisticRegression
+from sklearn.preprocessing import MinMaxScaler
+
+from kale.pipeline.mida_trainer import MIDATrainer
+
+N_SAMPLES = 200
+
+
+@pytest.fixture(scope="module")
+def toy_data():
```
xianyuanliu, line 14: It would be better to introduce this toy data, including its dimensions. We can consider how to reuse it for other related methods.

tests/pipeline/test_mida_trainer.py
```diff
@@ -0,0 +1,60 @@
+import numpy as np
+import pytest
+from numpy import testing
+from sklearn.datasets import make_blobs
+from sklearn.linear_model import LogisticRegression
+from sklearn.preprocessing import MinMaxScaler
+
+from kale.pipeline.mida_trainer import MIDATrainer
+
+N_SAMPLES = 200
```
xianyuanliu, line 10: Is the value of 200 needed? Shall we reduce it, even though it might not save many resources?

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def decision_function(self, x, groups=None):
+        """Decision scores of each class for the given data x
+
+        Args:
+            x (array-like tensor): input data, shape (n_samples, n_features)
+            groups (array-like, optional): group labels for the samples, shape (n_samples,)
+
+        Returns:
+            array-like: decision scores, shape (n_samples,) for binary case, else (n_samples, n_class)
```
xianyuanliu, line 157: Please ensure consistency in capitalising the word "decision" throughout this file.

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def predict_proba(self, x, groups=None):
+        """Probability of each class for the given data x. Not supported by "linear_svc".
+
+        Args:
+            x (array-like tensor): input data, shape (n_samples, n_features)
+            groups (array-like, optional): group labels for the samples, shape (n_samples,)
+
+        Returns:
+            array-like: probabilities, shape (n_samples, n_class)
```
xianyuanliu, line 170: The same as the previous one.

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,200 @@
...
+            self.estimator_param_grid = estimator_param_grid
+        self.best_mida_params = None
+        self.best_transformer = None
+        self.best_transformer_params = None
+        self.best_estimator = None
+        self.best_estimator_params = None
+        self.best_score = None
+
+    def fit(self, x, y, groups=None):
+        """Fit a pipeline with the given data x and labels y
```
xianyuanliu, line 66: What does "a pipeline" refer to? Please add more details. This line appears many times in the following different functions. Please refine them as well.

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,200 @@
...
+            x (array-like tensor): input data, shape (n_samples, n_features)
+            groups (array-like, optional): group labels for the samples, shape (n_samples,)
+
+        Returns:
+            array-like: probabilities, shape (n_samples, n_class)
+        """
+        return self.best_estimator.predict_proba(self.transform(x, groups=groups))
+
+    def transform(self, x, groups=None):
+        """Fit a pipeline with the given data x and labels y
```
xianyuanliu, line 175: Is this "a pipeline" or "transformer"? Please add more details.

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,200 @@
...
+            groups (array-like, optional): group labels for the samples, shape (n_samples,)
+
+        Returns:
+            array-like: Transformed data, shape (n_samples, n_features)
+        """
+
+        return self.best_mida.transform(self.best_transformer.transform(x), groups=groups)
+
+    def fit_predict(self, x, y, groups=None):
+        """Fit a pipeline with the given data x and labels y
```
xianyuanliu, line 189: This introduction is unclear. Please rewrite it by adding more details.

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,200 @@
...
+            self.mida_param_grid = mida_param_grid
+        if transformer_param_grid is None:
+            self.transformer_param_grid = {key: [val] for key, val in self.transformer.get_params().items()}
+        else:
+            self.transformer_param_grid = transformer_param_grid
+        if estimator_param_grid is None:
+            self.estimator_param_grid = {key: [val] for key, val in self.estimator.get_params().items()}
+        else:
+            self.estimator_param_grid = estimator_param_grid
+        self.best_mida_params = None
```
xianyuanliu, line 58: Is using a dictionary to save and update these parameters better and more compact?  We can talk about it. 
```suggestion
        self.best_params = {
              'mida_params': None,
              'transformer': None,
              'transformer_params': None,
              'estimator': None,
              'estimator_params': None,
              'score': None
}
```

xianyuanliu (changes requested):
Well done. I have successfully rerun the tutorial using Colab. The comments and some code structures could be improved. Please note that Codecov reminds us that some codes in `kale/embed/factorization.py` have not been tested.

kale/embed/factorization.py
```diff
@@ -301,7 +302,9 @@ def __init__(
             self.kernel_params = kernel_params
         self._label_binarizer = LabelBinarizer(pos_label=1, neg_label=-1)
         self._centerer = KernelCenterer()
+        self.fit_label = fit_label
         self.x_fit = None
+        self.enc = OneHotEncoder(handle_unknown="ignore")
```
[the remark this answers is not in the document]
zaRizk7: Another thing about scikit-learn's estimator is that you need to define only the arguments for `__init__` inside the function so if you only have for example like mu and eta, you only have to set those specifically. this is due to how it interacts with the `set_params(...)` methods, hence you cannot add or initiate other variables excluding the ones defined in `__init__` arguments.

kale/embed/factorization.py
```diff
@@ -301,7 +301,9 @@ def __init__(
...
 
     def _get_kernel(self, x, y=None):
         if self.kernel in ["linear", "rbf", "poly"]:
```
zaRizk7, line 310: rather than raising the error manually, I would recommend to exploit the _parameter_constraints implementation from sklearn to validate whether the argument is valid or not, you can refer to the `BaseEstimator` source code for that.

zaRizk7: here's a small example on how to define `_parameter_constraints` in the estimator

```py
class LogisticRegression(LinearClassifierMixin, SparseCoefMixin, BaseEstimator):

    _parameter_constraints: dict = {
            "penalty": [StrOptions({"l1", "l2", "elasticnet"}), None],
            "dual": ["boolean"],
            "tol": [Interval(Real, 0, None, closed="left")],
            "C": [Interval(Real, 0, None, closed="right")],
            "fit_intercept": ["boolean"],
            "intercept_scaling": [Interval(Real, 0, None, closed="neither")],
            "class_weight": [dict, StrOptions({"balanced"}), None],
            "random_state": ["random_state"],
    }
```

while its not on the official documentation of `sklearn`, you may refer to the source code in `sklearn/utils/_param_validation.py`

kale/embed/factorization.py
```diff
@@ -310,37 +312,43 @@ def _get_kernel(self, x, y=None):
             raise ValueError("Pre-computed kernel not supported")
         return pairwise_kernels(x, y, metric=self.kernel, filter_params=True, **params)
```
zaRizk7, line 314: To make lives easier, I would actually recommend to just let the argument be a callable rather than directly a function. so in `__init__`

zaRizk7: or if it is desirable to let it remain like that, maybe allow `__init__` to take all of the supported kernel arguments like `coef0` or `gamma`

kale/embed/factorization.py
```diff
@@ -301,7 +301,9 @@ def __init__(
             self.kernel_params = kernel_params
         self._label_binarizer = LabelBinarizer(pos_label=1, neg_label=-1)
         self._centerer = KernelCenterer()
+        self.fit_label = fit_label
         self.x_fit = None
+        self.enc = OneHotEncoder(handle_unknown="ignore")
```
zaRizk7, line 307: In this case, I would prefer to put this one inside the `fit` method rather than `__init__`. I would also recommend adding the underscore postfix (e.g., `enc_`) as it is like the default way to define additional attributes defined during fit in sklearn.

kale/embed/factorization.py
```diff
@@ -386,31 +394,32 @@ def _fit(self, kernel_x, y, covariates=None):
...
         """
         Args:
             x : array-like, shape (n_samples, n_features)
-            covariates : array-like, augmentation features, shape (n_samples, n_covariates)
+            groups : array-like, augmentation features, shape (n_samples, )
         Returns:
             x_transformed : array-like, shape (n_samples, n_components)
         """
         check_is_fitted(self, "x_fit")
-        if isinstance(covariates, np.ndarray) and self.augmentation:
```
zaRizk7, line 412: before doing anything with the input, I would recommend using the `BaseEstimator`'s `_validate_data` method which has the same arguments as `check_X_y` and `check_array` but the additional thing it has is that it also checks whether or not the input array has the expected shape. An example could be:

```py
def fit(self, X, y=None):
    X = self._validate_data(X, **other_check_array_argument)
    # do stuff
    return self

def transform(self, X):
    X = self._validate_data(X, reset=False)
    # Do stuff
    return X
```

It is necessary to set reset to `False` in `transform` as it means that the input size is already fixed.

zaRizk7: However if `covariates` or `groups` is already defined, you could use an additional `check_array` and `check_consistent_length`, to ensure the input is valid

zaRizk7: In this case, I would recommend to also override `_validate_data` to accommodate `groups` or `covariates`

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,200 @@
+# =============================================================================
+# Author: Shuo Zhou, shuo.zhou@sheffield.ac.uk
+# =============================================================================
+
+"""Implementation of MIDA->Feature Selection->Linear SVM/LogisticRegression Pipeline
```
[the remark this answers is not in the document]
zaRizk7: Does it makes sense that whether it can be generalized to something like `UDAGridSearchCV` or `DAGridSearchCV`?  where we note that `DA` and `UDA` imply domain adaptation and unsupervised domain adaptation specifically.

kale/embed/factorization.py
```diff
@@ -310,37 +312,43 @@ def _get_kernel(self, x, y=None):
...
+            x = np.concatenate((x, groups), axis=1)
+
+        groups = self.enc.fit_transform(groups.reshape(-1, 1)).toarray()
 
         # Kernel matrix
         kernel_x = self._get_kernel(x)
         kernel_x[np.isnan(kernel_x)] = 0
 
+        if self.n_components is None:
+            self.n_components = kernel_x.shape[0]
```
zaRizk7, line 337: recommend to add a warning in case if the kernel matrix's size is too large to do eigendecomposition. also, try to avoid mutating the defined attributes from `__init__` as it will causes an issue with `set_params(...)`. if `n_components` is too large, don't change it, rather just make a warning and then compute the eigenvalues with supported dimension size.

examples/cmri_mpca/tutorial.ipynb
```diff
@@ -1,405 +1,405 @@
...
+    "    - Registration\n",
+    "    - Masking\n",
+    "    - Rescaling\n",
+    "    - Normalization\n",
+    "- Machine learning pipeline:\n",
+    "    - Multilinear principal component analysis\n",
+    "    - Discriminative feature selection\n",
+    "    - Linear classification model training    \n",
+    "\n",
+    "[See the full pipeline illustration for more details.](https://oup.silverchair-cdn.com/oup/backfile/Content_public/Journal/ehjcimaging/22/2/10.1093_ehjci_jeaa001/2/m_jeaa001f2.jpeg?Expires=1631272906&Signature=PKl6KLDSoNyiTy~GNtXayJCucGKhweXvGz~svHY~ThjqjbokMVCnyJMMjoGQ4C81HdUcdnJA-rcNaqmDjspUs5eAX7avG~ckkIGXqGbPWrfnaMfwywWG3EXvvH0tw9ZcFeelnWgF4lkT0RFDVgzvzhHBvefNgS0ZGwLqiGJduANJFmWIXvYgNiU6M6kRbdpOJZBltknUO~Jv43-ghqmEX7dTfOKAx6~14quDC5cgzVFfwNFRKSn0P5JZRf~wIhQ6GQ4tprl7eXuzuRHcKnFbM~UkZOtcQvVhJofCCgSDnExyS6bns9Dop39OlfQHUdY4cwn1WaSnMKEqAqQaKZ715w__&Key-Pair-Id=APKAIE5G5CRDK6RD3PGA)\n",
```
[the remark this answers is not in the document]
shuo-zhou: Deleted and use the reference below instead

kale/embed/factorization.py
```diff
@@ -282,11 +283,11 @@ class MIDA(BaseEstimator, TransformerMixin):
 
     def __init__(
         self,
-        n_components,
+        n_components=None,
```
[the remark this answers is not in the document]
shuo-zhou: We follow the `KernelPCA` implementation in [Scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/70fdc843a/sklearn/decomposition/_kernel_pca.py#L32)

kale/embed/factorization.py
```diff
@@ -310,37 +313,42 @@ def _get_kernel(self, x, y=None):
...
-    def _fit(self, kernel_x, y, covariates=None):
+    def _fit(self, kernel_x, y=None, groups=None):
         """solve MIDA
 
         Args:
             kernel_x: array-like, kernel matrix of input data x, shape (n_samples, n_samples)
-            y: array-like. Labels, shape (nl_samples,)
-            covariates: array-like. Domain co-variates, shape (n_samples, n_covariates)
+            y: array-like. Labels, shape (n_labeled_samples,)
+            groups: array-like. Domain co-variates, shape (n_samples + nt_samples, n_groups)
```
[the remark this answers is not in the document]
shuo-zhou: Onehot matrix

kale/embed/factorization.py
```diff
@@ -352,12 +360,12 @@ def _fit(self, kernel_x, y, covariates=None):
...
         kernel_x = self._centerer.fit_transform(kernel_x)
-        if isinstance(covariates, np.ndarray):
-            kernel_c = np.dot(covariates, covariates.T)
+        if isinstance(groups, np.ndarray):
+            kernel_c = np.dot(groups, groups.T)
         else:
             kernel_c = np.zeros((n_samples, n_samples))
 
-        if y is not None:
+        if y is not None and self.fit_label:
```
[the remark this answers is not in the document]
shuo-zhou: updated docstring

kale/embed/factorization.py
```diff
@@ -310,37 +313,42 @@ def _get_kernel(self, x, y=None):
...
 
-    def fit(self, x, y=None, covariates=None):
+    def fit(self, x, y=None, groups=None):
         """
         Args:
             x : array-like. Input data, shape (n_samples, n_features)
-            y : array-like. Labels, shape (nl_samples,)
-            covariates : array-like. Domain co-variates, shape (n_samples, n_co-variates)
+            y : array-like. Labels, shape (n_labeled_samples,)
+            groups : array-like. Domain labels/grouping factors, shape (n_samples, )
```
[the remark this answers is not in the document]
shuo-zhou: for consistency with the variable name [here](https://github.com/pykale/pykale/blob/main/kale/evaluate/cross_validation.py#L11)

kale/embed/factorization.py
shuo-zhou: following the `KernelPCA` implementation in scikit-learn

kale/embed/factorization.py
shuo-zhou: updated in _fit()

kale/pipeline/mida_trainer.py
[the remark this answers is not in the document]
shuo-zhou: Might be better to keep consistency with other pipelines in Pykale

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,200 @@
...
+        transformer_param_grid (dict, optional): Grids for searching the optimal hyper-parameters of the transformer.
+            Defaults to None.
+        estimator_param_grid (dict, optional): Grids for searching the optimal hyper-parameters of the estimator.
+            Defaults to None.
+    """
+
+    def __init__(
+        self,
+        estimator,
+        estimator_param_grid,
```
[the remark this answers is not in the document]
shuo-zhou: `estimator_param_grid` is required, not optional, docstring has been updated

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+class MIDATrainer(BaseEstimator, ClassifierMixin):
+    """Trainer of pipeline: Transformer (optional) -> Maximum independence domain adaptation -> Estimator
+
+    Args:
+        estimator (BaseEstimator): Estimator object implementing 'fit'
+        estimator_param_grid (dict, optional): Grids for searching the optimal hyperparameters of the estimator.
+            Defaults to None.
+        transformer (BaseEstimator, optional): Transformer object implementing 'fit_transform'. Defaults to None.
```
zaRizk7, line 26: need clarification (more explicit definition) on what the `transformer` actually does? one idea to rename it might be `pre_mida_transformer` or another more concise naming?

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,201 @@
...
+        transformer=None,
+        mida_param_grid=None,
+        transformer_param_grid=None,
+    ):
+        self.estimator = estimator
+        self.transformer = transformer
+        self.mida_param_grid = mida_param_grid
+        self.transformer_param_grid = transformer_param_grid
+        self.estimator_param_grid = estimator_param_grid
+        self.best_transformer = None
```
zaRizk7, line 45: initialize the `best` attributes during `fit` rather than `__init__`. and also use standard way on how to define attributes during fit such as `best_mida_`

kale/embed/factorization.py
```diff
@@ -267,80 +269,123 @@ def inverse_transform(self, x):
...
             params = self.kernel_params or {}
         else:
-            raise ValueError("Pre-computed kernel not supported")
-        return pairwise_kernels(x, y, metric=self.kernel, filter_params=True, **params)
+            self.gamma = 1 / x.shape[1] if self.gamma is None else self.gamma
+            params = {"gamma": self.gamma, "degree": self.degree, "coef0": self.coef0}
+        return pairwise_kernels(x, y, metric=self.kernel, filter_params=True, n_jobs=self.n_jobs, **params)
 
-    def fit(self, x, y=None, covariates=None):
+    def fit(self, x, y=None, groups=None):
```
zaRizk7, line 352: To make `_parameter_constraints` work, don't forget to add `@_fit_context(prefer_skip_nested_validation=True)` on top of `fit` as shown by various `sklearn` estimator's source code.

zaRizk7: the function could be imported from `sklearn.base`

kale/embed/factorization.py
```diff
@@ -267,80 +269,123 @@ def inverse_transform(self, x):
...
+        self.degree = degree
+        self.coef0 = coef0
         self.mu = mu
         self.eta = eta
         self.augmentation = augmentation
-        if kernel_params is None:
-            self.kernel_params = {}
-        else:
-            self.kernel_params = kernel_params
-        self._label_binarizer = LabelBinarizer(pos_label=1, neg_label=-1)
```
zaRizk7, line 302: are there any reason why `neg_label=-1`? because the original paper by (Yan et al., 2017) states:

image

zaRizk7: I think assuming that having no label implies that `y_i=-1` might be one idea to achieve similar outcome from the original paper. An example where this assumption has been used is from `sklearn.semi_supervised.LabelPropagation`.

zaRizk7: we could then get the indices for labeled and unlabeled data to augment the one-hot encoded/binarized representation where the unlabeled data will only have zeros as values.

kale/embed/factorization.py
```diff
@@ -267,80 +269,123 @@ def inverse_transform(self, x):
...
+            callable,
+        ],
+        "gamma": [
+            Interval(Real, 0, None, closed="left"),
+            None,
+        ],
+        "degree": [Interval(Real, 0, None, closed="left")],
+        "coef0": [Interval(Real, None, None, closed="neither")],
+        "kernel_params": [dict, None],
+        "mu": [Real],
```
zaRizk7, line 307: constraint both `mu` and `eta` to be strictly positive like in the original paper (Yan et al., 2017).

image

`eta=gamma` in this case
image

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,201 @@
...
+        self.best_estimator.fit(x_transformed[y.shape[0] :,], y)
+
+        return self
+
+    def _fit_mida(self, x_transformed, y, groups, transformer_params):
+        """Fit MIDA and estimator with the given data x_transformed and labels
+
+        Args:
+            x_transformed (array-like): Transformed data, shape (n_samples, n_features)
+            y (array-like): Data labels, shape (n_labeled_samples,)
```
zaRizk7, line 90: I think just allowing unlabeled `y` to be -1 is a preferable option. we could easily index the labeled data when the label and unlabeled data is in a random order.

zaRizk7: it will make our assumptions easier when doing `_validate_data` as it expects `X` and `y` to have the same number of samples.

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,201 @@
...
+        clf.fit(x_transformed[y.shape[0] :,], y)
+        if self.best_score is None or clf.best_score_ > self.best_score:
+            self.best_score = clf.best_score_
+            self.best_estimator_params = clf.best_params_
+            self.best_mida_params = mida_params
+            self.best_transformer_params = transformer_params
+
+        return self
+
+    def predict(self, x, groups=None):
```
zaRizk7, line 136: This could be useful to add to handle `predict`, `transform`, `predict_proba`, etc. 

`available_if`: https://scikit-learn.org/stable/modules/generated/sklearn.utils.metaestimators.available_if.html

kale/embed/factorization.py
```diff
@@ -267,80 +269,123 @@ def inverse_transform(self, x):
...
+            y (array-like): Labels, shape (n_labeled_samples,)
+            groups (array-like): Domain/group labels, shape (n_samples,)
 
         Note:
-            Unsupervised MIDA is performed if y is None.
-            Semi-supervised MIDA is performed is y is not None.
+            Unsupervised MIDA is performed if y is None or fit_label=False.
+            Semi-supervised MIDA is performed is y is not None and fit_label=True.
         """
-        if self.augmentation and isinstance(covariates, np.ndarray):
```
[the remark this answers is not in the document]
zaRizk7: By overriding `_validate_data` we could also do a check whether the group during `transform` is expected and is defined during fit or not.

kale/embed/factorization.py
zaRizk7: in return, we could avoid an assumption where the data must be ordered from labeled to unlabeled data and we will still be able to do `_validate_data` check properly.

kale/embed/factorization.py
```diff
@@ -267,80 +269,123 @@ def inverse_transform(self, x):
...
-        else:
-            self.kernel_params = kernel_params
-        self._label_binarizer = LabelBinarizer(pos_label=1, neg_label=-1)
-        self._centerer = KernelCenterer()
+        self.kernel_params = kernel_params
+        self.fit_label = fit_label
         self.x_fit = None
+        self._label_binarizer = None
+        self._centerer = None
+        self._onehot = None
```
zaRizk7, line 341: Likewise for `_label_binarizer`, `_centerer`, and `_onehot`, we only initialize it during `fit`. And use standard definition of attributes initialized during `fit` such as `centerer_`. I would recommend these to be moved to avoid messing with `BaseEstimator.set_params(...)`

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,201 @@
...
+            self.transformer = FunctionTransformer()
+        if self.transformer_param_grid is None:
+            self.transformer_param_grid = {key: [val] for key, val in self.transformer.get_params().items()}
+        transformer_grid = ParameterGrid(self.transformer_param_grid)
+        for transformer_params in transformer_grid:
+            self.transformer.set_params(**transformer_params)
+            x_transformed = self.transformer.fit_transform(x)
+            self._fit_mida(x_transformed, y, groups, transformer_params)
+
+        self.best_transformer = self.transformer.set_params(**self.best_transformer_params)
```
zaRizk7, line 75: I think cloning the original estimator with `sklearn.base.clone` first and then do `set_params` could be a better option to avoid messing with the original estimator.

shuo-zhou (commented):
The raise error lines should be easy to cover. Let me know if you need help.

kale/embed/factorization.py
```diff
@@ -63,6 +76,225 @@ def _check_tensor_dim_shape(x, n_dims, shape_):
     _check_shape(x, shape_)
 
 
+def _centering_kernel(size, dtype=np.float64):
```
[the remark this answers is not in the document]
zaRizk7: Uunfortunately, `scikit-learn` they don't have an explicit implementation to produce a centering kernel, they do it implicitly through `KernelCenterer` since they would like to support centering during inference where they don't need to store the kernel centered, hence we need the custom function for centering kernel

The solvers are originally derived from `scikit-learn`'s implementation for hermitian eigendecomposition for symmetric matrices which `KernelPCA` uses

kale/embed/factorization.py
```diff
@@ -63,6 +76,225 @@ def _check_tensor_dim_shape(x, n_dims, shape_):
...
+            b,
+            which="LA",
+            tol=tol,
+            maxiter=max_iter,
+            v0=v0,
+        )
+
+    if solver == "randomized":
+        if b is not None:
+            a = la.inv(b) @ a
```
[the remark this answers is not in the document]
zaRizk7: should we add a test case separately for internal functions then?

kale/pipeline/mida_trainer.py
```diff
@@ -0,0 +1,451 @@
...
+        return self.best_mida_.n_features_in_
+
+    @_fit_context(prefer_skip_nested_validation=True)
+    def fit(self, x, y=None, factors=None, **params):
+        """Fit and tune the pipeline with the given data.
+
+        Args:
+            x (array-like): The input data.
+            y (array-like): The target values.
+            factors (array-like, optional): The factors influencing the input data.
```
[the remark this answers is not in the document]
zaRizk7: One reason why the `factors` is not a domain label is that we would like to incorporate multiple domain information which may not be captured if we just specify groups. In some use cases, we may want to incorporate multiple information which may factor in the distributional shift of the dataset to integrate the dataset more adequately. If we only consider a single domain label as a factor, then we can treat `groups` = `domains` in this case.

I will update the doc to make it clearer by stating that the domain information need to be preprocessed (e.g. doing one-hot) before using `MIDA`

kale/embed/factorization.py
```diff
@@ -265,154 +497,460 @@ def inverse_transform(self, x):
...
+    def _make_solution_kernel(self, k_x, y, factors):
+        h = _centering_kernel(_num_features(k_x), k_x.dtype)
+        k_y = pairwise_kernels(y, n_jobs=self.num_jobs)
+        k_f = pairwise_kernels(factors, n_jobs=self.num_jobs)
+
+        centerer = KernelCenterer()
+        k_y = centerer.fit_transform(k_y)
+        k_f = centerer.fit_transform(k_f)
+
+        k_solution = multi_dot((k_x, self.mu * h + self.eta * k_y - k_f, k_x))
```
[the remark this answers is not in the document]
zaRizk7: yes, i concur

kale/embed/factorization.py
```diff
@@ -63,6 +76,225 @@ def _check_tensor_dim_shape(x, n_dims, shape_):
...
+
+# Postprocess eignevalues and eigenvectors
+
+
+def _postprocess_eigencomponents(eigenvalues, eigenvectors, steps, num_components=None, remove_zero_eig=False):
+    """Postprocess the eigenvalues and eigenvectors after eigendecomposition.
+    Args:
+        eigenvalues (array-like): The eigenvalues of the kernel matrix.
+        eigenvectors (array-like): The eigenvectors of the kernel matrix.
+        steps (list): The steps to perform in postprocessing.
```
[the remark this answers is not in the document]
zaRizk7: Added available options as requested

xianyuanliu (approved):
@shuo-zhou create a new branch to refactor further and improve test coverage

PR #511, opened 2025-12-28
Fix torch-numpy type mismatch

haipinglu:
### Description
- **Motivation:** PyKale’s CI was failing because `sklearn`’s `LabelBinarizer.inverse_transform` was receiving a torch tensor from `CoIRLS.predict()` in kale/pipeline/multi_domain_adapter.py, which `sklearn` does not accept.
- **Summary:** Always detach tensor outputs to CPU NumPy arrays before sending them to `inverse_transform`, simplifying the binary path while keeping behavior unchanged otherwise.
- **Tests:** Tested selected code locally. CI passed.

### Status
**Ready**

### Types of changes
- [ ] Source for documentation at docs manually updated for new API.

### DIFF
--- a/kale/pipeline/multi_domain_adapter.py
+++ b/kale/pipeline/multi_domain_adapter.py
@@ -439,9 +439,9 @@ class CoIRLS(BaseEstimator, ClassifierMixin):
         """
         out = self.decision_function(x)
         if self._label_binarizer.y_type_ == "binary":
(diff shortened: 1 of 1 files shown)


xianyuanliu (approved):
LGTM. One minor concern is whether this device switching reduces speed. It works for now, and we can address it if needed later.

PR #528, opened 2026-04-17
CI: prevent changelog bot from overwriting manual edits

cmanjah:
## Problem

The changelog bot triggers on `synchronize` events (every push to the PR branch). Since the bot itself commits back to the branch (`commit_changelog: true`), this causes:
1. Every push overwrites any manual changelog edits
2. The bot's own commit re-triggers itself (loop risk)

## Fix

- Remove `synchronize` from the trigger types
- Add `paths-ignore: ['.github/CHANGELOG.md']` so bot commits don't re-trigger the workflow

## Impact

The bot now only runs once when a release PR is opened/reopened/edited — not on every push.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -5,7 +5,7 @@ name: changelog
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, edited]
+    types: [opened]
 
 jobs:
   log-changes:


.github/workflows/changelog.yml
```diff
@@ -5,7 +5,9 @@ name: changelog
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, edited]
+    types: [opened, reopened, edited]
```
haipinglu, line 8: I think the trigger should be just "opened, reopened", isn't it?
Actually, I prefer a fresh PR (so I keep only "opened" as the trigger previously). Now the commit messages are very noisy. You should be aware that all commits will be merged into the main. If you agree this is better, then no need to add "reopened".

cmanjah: Agreed — opened only is cleaner. Updated the trigger to `types: [opened]` and also removed the `paths-ignore: [".github/CHANGELOG.md"]` guard, which was only needed to prevent a re-trigger loop from the bot's own commit back to the branch.

haipinglu (approved):

PR #574, opened 2026-07-15
Fix cross-DataFrame index alignment in bin_wise_bound_eval and bin_wise_errors

### LINKED ISSUE #553: `[Bug]` Cross-DataFrame boolean indexing of `fold_bins` with `fold_errors` mask is alignment-fragile
# 🐛 Bug

In both `bin_wise_bound_eval` (`:1631`) and `bin_wise_errors` (`:1911`), `fold_bins` is filtered using a
boolean mask built from a *different* DataFrame:

```python
pred_bins_ti = fold_bins[(fold_errors["Target Index"] == target_idx)][...]
```

This only produces correct rows if `fold_bins` and `fold_errors` share identical, aligned indices. If the
two frames are ever reindexed/reset independently, pandas aligns by index and silently returns wrong or
`NaN` rows.

## Expected Behaviour
Filter `fold_bins` on its own `"Target Index"` column (`fold_bins[fold_bins["Target Index"] == target_idx]`),
or merge the frames on `uid` first.

## Environment

```
python collect_env.py
```

 - OS (e.g., Linux):
 - Python version:
 - CUDA/cuDNN version:

## Additional context
Latent correctness risk; surfaced while reviewing for #410. Easy to fix during the evaluator refactor.

cmanjah:
Fixes #553.

### Description
`bin_wise_bound_eval` and `bin_wise_errors` filtered `fold_bins` using a boolean mask built from a *different* DataFrame (`fold_errors["Target Index"] == target_idx`). This relied on `fold_errors` and `fold_bins` always sharing an identical index and row order. That invariant holds for the current callers — both frames are sibling slices of the same `data_structs` taken with the same `Testing Fold == fold` mask — but it is implicit and fragile: if either frame is ever reindexed independently, pandas aligns by index and silently returns the wrong rows.

Both functions now filter `fold_bins` on its own `Target Index` column, removing the hidden cross-DataFrame alignment dependency. Because the two frames are aligned today, the change is behavior-preserving for existing callers; it hardens against a latent, silent mis-grouping bug.

Adds `TestCrossDataFrameIndexAlignment`, which deliberately reorders/reindexes `fold_bins` and asserts that grouping is still by `uid`:
- `test_bin_wise_errors_misaligned_index` — checks a hand-computed expectation and equality with the aligned result.
- `test_bin_wise_bound_eval_misaligned_index` — checks bound accuracy is identical for aligned vs misaligned frames.

Both tests fail on the previous cross-frame indexing and pass with this fix.

### Status
Ready

### Types of changes

### DIFF
--- a/kale/evaluate/uncertainty_metrics.py
+++ b/kale/evaluate/uncertainty_metrics.py
@@ -23,7 +23,6 @@ Main Classes:
     - QuantileCalculator: Quantile-based error distribution analysis
     - MetricsCalculator: Statistical metrics computation
 """
-
 import copy
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


xianyuanliu (approved):
LGTM

PR #578, opened 2026-07-15
Use the model device instead of hard-coded .cuda() in SignalImageVAE

### LINKED ISSUE #542: `[Bug]` Hard-coded `.cuda()` breaks device-agnostic execution in `MultimodalVAE`
# 🐛 Bug

```python
# kale/embed/multimodal_encoder.py:76-77
if use_cuda:
    mean = mean.cuda()
    log_var = log_var.cuda()
```

`.cuda()` forces device 0 and ignores the model's actual device, causing device-mismatch errors on
multi-GPU setups, a specific CUDA index, or `mps`.

## To reproduce

1. Place the model on a non-default device (e.g. `cuda:1` or `mps`).
2. Run a forward pass that calls this helper.
3. A device-mismatch `RuntimeError` is raised.

## Expected Behaviour

Tensors are created on the same device as the module (e.g. via `.to(device)` or by deriving the device
from an existing parameter).

## Environment

```
python collect_env.py
```

 - OS (e.g., Linux):
 - Python version:
 - CUDA/cuDNN version:

## Additional context
Fix: thread a `device` argument through and use `.to(device)`.

cmanjah:
Fixes #542.

### Description
`SignalImageVAE.prior_expert` used `.cuda()`, which forces device 0 and ignores the model's actual device. `infer()` now derives the device from the model parameters and creates the prior there, so it works on non-default GPUs and MPS. Adds device-placement and forwarding tests.

### Status
Ready

### Types of changes

### DIFF
--- a/tests/embed/test_multimodal_encoder.py
+++ b/tests/embed/test_multimodal_encoder.py
@@ -29,7 +29,11 @@ def test_bimodal_vae_full_coverage():
         mu_cuda, log_var_cuda = model.prior_expert((1, batch_size, latent_dim), use_cuda=True)
         assert mu_cuda.shape == (1, batch_size, latent_dim)
         assert log_var_cuda.shape == (1, batch_size, latent_dim)
-        assert mock_zeros.call_args.kwargs["device"] == "cuda"
+        # Both the mean and log-variance tensors must be created on the cuda device, so check every
+        # call rather than only the last one - otherwise a regression that placed just one of them
+        # correctly would slip through.
+        assert mock_zeros.call_count == 2
+        assert all(call.kwargs["device"] == "cuda" for call in mock_zeros.call_args_list)
 
     # --- Test reparametrize, both training and eval mode ---
     dummy_mu = torch.zeros(batch_size, latent_dim)
--- a/tests/embed/test_multimodal_encoder.py
+++ b/tests/embed/test_multimodal_encoder.py
@@ -23,12 +23,13 @@ def test_bimodal_vae_full_coverage():
     assert mean.shape == (1, batch_size, latent_dim)
     assert torch.allclose(log_var, torch.zeros_like(log_var))
 
-    # --- Test prior_expert (mocked cuda branch, even on CPU) ---
-    with patch.object(torch.Tensor, "cuda", lambda x: x):
+    # --- Test prior_expert (cuda branch resolves to the cuda device, even on CPU-only hosts) ---
+    prior_stub = torch.zeros(1, batch_size, latent_dim)  # built before patching torch.zeros
+    with patch("kale.embed.multimodal_encoder.torch.zeros", return_value=prior_stub) as mock_zeros:
         mu_cuda, log_var_cuda = model.prior_expert((1, batch_size, latent_dim), use_cuda=True)
         assert mu_cuda.shape == (1, batch_size, latent_dim)
         assert log_var_cuda.shape == (1, batch_size, latent_dim)
-        # These are not .is_cuda (since we're faking), but they exist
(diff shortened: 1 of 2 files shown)


tests/embed/test_multimodal_encoder.py
```diff
@@ -23,12 +23,13 @@ def test_bimodal_vae_full_coverage():
...
-    # --- Test prior_expert (mocked cuda branch, even on CPU) ---
-    with patch.object(torch.Tensor, "cuda", lambda x: x):
+    # --- Test prior_expert (cuda branch resolves to the cuda device, even on CPU-only hosts) ---
+    prior_stub = torch.zeros(1, batch_size, latent_dim)  # built before patching torch.zeros
+    with patch("kale.embed.multimodal_encoder.torch.zeros", return_value=prior_stub) as mock_zeros:
         mu_cuda, log_var_cuda = model.prior_expert((1, batch_size, latent_dim), use_cuda=True)
         assert mu_cuda.shape == (1, batch_size, latent_dim)
         assert log_var_cuda.shape == (1, batch_size, latent_dim)
-        # These are not .is_cuda (since we're faking), but they exist
+        assert mock_zeros.call_args.kwargs["device"] == "cuda"
```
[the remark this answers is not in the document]
cmanjah: Fixed. The assertion now checks `mock_zeros.call_count == 2` and that **every** call in `call_args_list` used `device == "cuda"`, rather than only the last call.

Confirmed it was a real gap, not a hypothetical: I injected a partial bug (placing `log_var` on the device but leaving `mean` on the default) and the old single-`call_args` assertion still passed; the strengthened one fails on it and passes on the correct code.

tests/embed/test_multimodal_encoder.py
```diff
@@ -53,3 +58,55 @@ def test_bimodal_vae_full_coverage():
...
+
+
+class TestPriorExpertDevice:
+    """Device placement of the prior expert (issue #542)."""
+
+    def test_explicit_device_is_honoured(self):
+        """An explicit device is applied regardless of the use_cuda flag."""
+        mean, log_var = SignalImageVAE.prior_expert((1, 2, 4), device="cpu")
+
+        assert mean.device.type == "cpu"
```
shuo-zhou, line 70: Can we only test GPU configurations locally?

shuo-zhou (commented):
Same as my [comments](https://github.com/pykale/pykale/pull/577#pullrequestreview-4748318906) in PR #577: how do other libraries test GPU usage?

tests/embed/test_multimodal_encoder.py
cmanjah: For now, yes — GPU execution can't run in CI here (`ubuntu-latest` has no GPU and `test.yml` installs the CPU torch build), so real-hardware runs are local. This test doesn't need a GPU: it checks that `infer()` forwards the model's own device to `prior_expert` (the #542 bug), which is verifiable on CPU.

On how other libraries handle it: the common pattern is to mark GPU tests so they auto-skip when `torch.cuda.is_available()` is False — e.g., PyTorch Lightning's `@RunIf(min_cuda_gpus=1)`, MONAI's `@skip_if_no_cuda`, HF's `@require_torch_gpu` — keep standard CI on CPU, and run the GPU-marked tests on a self-hosted or cloud GPU runner (nightly/dedicated); GitHub's hosted runners have none. We could adopt the same `@pytest.mark.gpu` skip pattern here, running those tests locally until/unless a GPU runner is set up.

shuo-zhou (approved):
Look good to me. A discussion about GPU testing related to this PR has been created here: https://github.com/pykale/pykale/discussions/589#discussion-10477851

PR #579, opened 2026-07-15
Handle optimizers without a scheduler in WDGRLTrainer.configure_optimizers

### LINKED ISSUE #548: [Bug] WDGRLTrainer crashes with ValueError when using default Adam optimizer
# 🐛 Bug
## Bug

`WDGRLTrainer.configure_optimizers()` crashes with `ValueError: not enough 
values to unpack (expected 2, got 1)` when using the default Adam optimizer 
(i.e. no explicit `optimizer` argument passed). Related to #544.

## Code to reproduce

```python
import torch
from unittest.mock import MagicMock
from torch.utils.data import DataLoader, TensorDataset
from kale.pipeline.domain_adapter import WDGRLTrainer, Method

fake_loader = DataLoader(
    TensorDataset(torch.randn(20, 10), torch.randint(0, 2, (20,))),
    batch_size=4
)
mock_dataset = MagicMock()
mock_dataset.prepare_data_loaders.return_value = (fake_loader, fake_loader)

m = WDGRLTrainer(
    dataset=mock_dataset,
    feature_extractor=torch.nn.Linear(10, 8),
    task_classifier=torch.nn.Linear(8, 2),
    critic=torch.nn.Linear(8, 1),
    method=Method.WDGRL
    # No optimizer arg — defaults to Adam
)
m.configure_optimizers()
```

## Root cause

`_configure_optimizer()` (`domain_adapter.py:411-450`) returns `[optimizer]` 
for Adam but `([optimizer], [scheduler])` for SGD.
`WDGRLTrainer.configure_optimizers()` (line 784) always unpacks two values, 
so it crashes with any non-SGD optimizer.

The tests in `test_domain_adapter.py` never catch this because every 
fixture hardcodes `optimizer={"type": "SGD", ...}`.

## Expected behaviour

`WDGRLTrainer` should work with Adam (the library default), consistent 
with all other trainers.

## Environment

- PyKale: 0.2.0
- PyTorch: 2.6.0+cpu
- PyTorch Lightning: 2.6.5
- Python: 3.11.9
- OS: Windows 11, Git Bash
- GPU: CPU only

cmanjah:
Fixes #548.

### Description
`WDGRLTrainer.configure_optimizers` assumed `_configure_optimizer` always returned `(optimizers, schedulers)`, so the default Adam optimizer (which returns just `[optimizer]`) raised `ValueError`. It now handles both shapes and passes the task result straight through. Adds tests for the Adam/AdamW and SGD paths.

### Status
Ready

### Types of changes

### DIFF
--- a/kale/pipeline/domain_adapter.py
+++ b/kale/pipeline/domain_adapter.py
@@ -784,8 +784,9 @@ class WDGRLTrainer(BaseDANNLike):
     def configure_optimizers(self):
         """Configure the task/feature optimizer and, separately, the critic optimizer.
 
-        The critic is stepped manually in ``compute_loss``, so its optimizer and scheduler are
-        stored on the trainer rather than returned to PyTorch Lightning.
+        The critic is stepped manually in ``critic_update_steps`` (called from ``training_step``),
+        so its optimizer and scheduler are stored on the trainer rather than returned to PyTorch
+        Lightning.
 
         ``_configure_optimizer`` returns ``([optimizer], [scheduler])`` only for SGD with
         ``adapt_lr`` enabled, and ``[optimizer]`` otherwise; both shapes are handled here and the
--- a/kale/pipeline/domain_adapter.py
+++ b/kale/pipeline/domain_adapter.py
@@ -781,24 +781,31 @@ class WDGRLTrainer(BaseDANNLike):
 
         return loss  # required, for backward pass
 
-    def configure_optimizers(self):  # no usage?
+    def configure_optimizers(self):
+        """Configure the task/feature optimizer and, separately, the critic optimizer.
+
+        The critic is stepped manually in ``compute_loss``, so its optimizer and scheduler are
+        stored on the trainer rather than returned to PyTorch Lightning.
+
+        ``_configure_optimizer`` returns ``([optimizer], [scheduler])`` only for SGD with
+        ``adapt_lr`` enabled, and ``[optimizer]`` otherwise; both shapes are handled here and the
+        task/feature result is passed straight through to Lightning.
+        """
         nets = [self.feat, self.classifier]
         parameters = set()
         for net in nets:
             parameters |= set(net.parameters())
 
-        if self._adapt_lr:
-            task_feat_optimizer, task_feat_sched = self._configure_optimizer(parameters)
-            self.critic_opt, self.critic_sched = self._configure_optimizer(self.domain_classifier.parameters())
-            self.critic_opt = self.critic_opt[0]
-            self.critic_sched = self.critic_sched[0]
-            return task_feat_optimizer, task_feat_sched
+        critic_result = self._configure_optimizer(self.domain_classifier.parameters())
+        if isinstance(critic_result, tuple):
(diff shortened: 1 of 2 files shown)


kale/pipeline/domain_adapter.py
```diff
@@ -781,24 +781,31 @@ def training_step(self, batch, batch_id):
 
         return loss  # required, for backward pass
 
-    def configure_optimizers(self):  # no usage?
+    def configure_optimizers(self):
+        """Configure the task/feature optimizer and, separately, the critic optimizer.
+
+        The critic is stepped manually in ``compute_loss``, so its optimizer and scheduler are
+        stored on the trainer rather than returned to PyTorch Lightning.
```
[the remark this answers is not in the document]
cmanjah: Fixed in d6d88690. Confirmed the reference: `self.critic_opt.step()` is in `critic_update_steps` (line 753), which `training_step` calls (line 762) — not `compute_loss`. The docstring now names `critic_update_steps` (called from `training_step`).

kale/pipeline/domain_adapter.py
```diff
@@ -781,24 +781,32 @@ def training_step(self, batch, batch_id):
...
             parameters |= set(net.parameters())
 
-        if self._adapt_lr:
-            task_feat_optimizer, task_feat_sched = self._configure_optimizer(parameters)
-            self.critic_opt, self.critic_sched = self._configure_optimizer(self.domain_classifier.parameters())
-            self.critic_opt = self.critic_opt[0]
-            self.critic_sched = self.critic_sched[0]
-            return task_feat_optimizer, task_feat_sched
+        critic_result = self._configure_optimizer(self.domain_classifier.parameters())
+        if isinstance(critic_result, tuple):
```
shuo-zhou, line 801: Would `if self._adapt_lr and isinstance(critic_result, tuple):` be better?

shuo-zhou (commented):
Please confirm whether `self._adapt_lr` need to be checked

kale/pipeline/domain_adapter.py
cmanjah: Thanks @shuo-zhou, I dug into this, and I don't think the extra check would actually change the behaviour.

`_configure_optimizer` only hands back a scheduler alongside the optimiser for SGD with `adapt_lr` on. For Adam, AdamW, and the default, it returns just the optimiser, whether or not `adapt_lr` is set. So any time `critic_result` is a tuple, `adapt_lr` is already true. In addition, it gives the same answer in all eight combinations of optimiser and `adapt_lr`.

I'd also rather keep `adapt_lr` out of this particular check. Assuming "`adapt_lr` is on, so there must be a scheduler" is what caused #548 in the first place the old `if self._adapt_lr:` branch tries to unpack two values and fails with `ValueError: not enough values to unpack (expected 2, got 1)` for Adam, AdamW, and the default. Going by what was actually returned is what stops that recurring: both new tests fail with that same error if I revert the fix.

The docstring just above the branch already spells out which shape comes back when, so I think the intent is covered.

PR #580, opened 2026-07-15
Load datasets with weights_only=True to prevent arbitrary code execution

### LINKED ISSUE #541: `[Bug]` Unsafe deserialization (`torch.load(weights_only=False)` / `pickle.load`) on downloaded data
# 🐛 Bug

Several dataset loaders deserialize remotely-fetched artifacts with mechanisms that allow arbitrary
code execution if the file is malicious or substituted:

- `kale/loaddata/polypharmacy_datasets.py:48` — `torch.load(data_path, weights_only=False)`
- `kale/loaddata/multiomics_datasets.py:77` — `torch.load(..., weights_only=False)`
- `kale/loaddata/mnistm.py:71,75` — `torch.load(...)` (no `weights_only`)
- `kale/loaddata/mnistm.py:150`, `kale/loaddata/usps.py:99` — `pickle.load(f, encoding="bytes")`

These files are downloaded via URL / Google Drive, so a compromised artifact is a real RCE vector.

## To reproduce
Code path: any dataset whose `processed` file is replaced with a crafted pickle/torch payload executes
embedded code on load.

## Expected Behaviour

Tensor payloads load with `weights_only=True`; where pickled Python objects are genuinely required, the
trust assumption is documented and downloads are checksum-verified.

## Environment

```
python collect_env.py
```

 - OS (e.g., Linux):
 - Python version:
 - CUDA/cuDNN version:

## Additional context
Pairs with the download-integrity request (no checksum verification currently). Fix: switch to
`weights_only=True` where possible; add hash verification for everything else.

cmanjah:
Fixes #541.

### Description
`torch.load(..., weights_only=False)` allowed arbitrary code execution from downloaded `.pt` files. MNIST-M moves to `weights_only=True` directly; the multiomics and polypharmacy loaders (PyTorch Geometric `Data`) go through a new `kale.utils.serialization.load_pyg_data`, which loads on the safe path while trusting only the container/array classes those objects need. Adds tests, including one that verifies a crafted payload is rejected rather than executed.

### Status
Ready

### Types of changes
- [x] Source for documentation at `docs` manually updated for new API.


kale/utils/serialization.py
```diff
@@ -0,0 +1,64 @@
...
+    from numpy._core.multiarray import _reconstruct
+    from torch_geometric.data.storage import GlobalStorage
+
+    allowed: List[Any] = [pyg_data.Data, GlobalStorage, _reconstruct, np.ndarray, np.dtype]
+
+    # The concrete dtype class depends on the stored array, so allow the whole family rather than
+    # chase them one at a time. They describe layout only and cannot execute code.
+    allowed.extend(
+        getattr(np_dtypes, name) for name in dir(np_dtypes) if isinstance(getattr(np_dtypes, name, None), type)
+    )
```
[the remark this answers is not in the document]
cmanjah: Fixed. The numpy internals (`_reconstruct` from `numpy._core.multiarray` and the `numpy.dtypes` family) are now resolved inside `try/except ImportError` blocks. `np.ndarray`/`np.dtype` are always allow-listed; a missing internal only means a particular payload cannot be rebuilt (a clear load-time error) rather than an ImportError that would break every loader importing this module.

kale/utils/serialization.py
```diff
@@ -0,0 +1,64 @@
...
+        **kwargs: Additional keyword arguments forwarded to :func:`torch.load`.
+
+    Returns:
+        The deserialized object, typically a ``Data`` instance or a list of them.
+
+    Example:
+        >>> data = load_pyg_data("data.pt")
+    """
+    with torch.serialization.safe_globals(get_pyg_safe_globals()):
+        return torch.load(path, weights_only=True, **kwargs)
```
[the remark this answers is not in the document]
cmanjah: Fixed. `load_pyg_data` now does `kwargs.pop("weights_only", None)` before `torch.load(path, weights_only=True, **kwargs)`, so a caller-supplied `weights_only` can no longer collide (no `TypeError`) or override the safe path. The docstring notes that `weights_only` is ignored.

tests/utils/test_serialization.py
```diff
@@ -0,0 +1,60 @@
...
+from torch_geometric.data import Data
+
+from kale.utils.serialization import get_pyg_safe_globals, load_pyg_data
+
+
+class Exploit:
+    """Stands in for a malicious payload; ``os.system`` would run on an unrestricted unpickle."""
+
+    def __reduce__(self):
+        return (eval, ("__import__('pathlib').Path(MARKER).write_text('executed')",))
```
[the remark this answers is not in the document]
cmanjah: Fixed. `Exploit` now takes a `marker_path` and embeds it into the payload string (`Path(r"...").write_text("executed")`), so if the payload ever executed it would leave an observable marker file rather than raising `NameError`.

tests/utils/test_serialization.py
```diff
@@ -0,0 +1,60 @@
...
+    container types needed by Data must not reopen the door to arbitrary code execution.
+    """
+    path = tmp_path / "payload.pt"
+    with open(path, "wb") as handle:
+        pickle.dump(Exploit(), handle)
+
+    with pytest.raises(Exception) as excinfo:
+        load_pyg_data(str(path))
+
+    assert "eval" in str(excinfo.value) or "Unsupported" in str(excinfo.value) or "Weights only" in str(excinfo.value)
```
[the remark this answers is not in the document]
cmanjah: Fixed, and this was the important one. The payload is now written with `torch.save`, so it travels the exact `torch.load` path the loader uses, and the test asserts `not marker.exists()`. Verified it has teeth: with the loader temporarily switched to `weights_only=False` the test fails (the payload executes / is not rejected); on the safe path it passes. So it can no longer false-pass on a file-format error.

tests/utils/test_serialization.py
```diff
@@ -0,0 +1,60 @@
+import pickle
+
+import pytest
```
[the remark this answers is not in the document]
cmanjah: Fixed - `pickle` is removed. flake8 (in the pinned pre-commit) is green, confirming no unused import remains.

kale/loaddata/pyg_serialization.py
```diff
@@ -0,0 +1,82 @@
...
+    # Present only in newer PyTorch Geometric releases, so resolve them defensively.
+    for name in ("DataEdgeAttr", "DataTensorAttr"):
+        attribute = getattr(pyg_data, name, None)
+        if attribute is not None:
+            allowed.append(attribute)
+
+    return allowed
+
+
+def load_pyg_data(path: str, **kwargs) -> Any:
```
shuo-zhou: It may be better if under the `loaddata` API

tests/loaddata/test_pyg_serialization.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    def __init__(self, marker_path):
+        self.marker_path = str(marker_path)
+
+    def __reduce__(self):
+        return (eval, (f"__import__('pathlib').Path(r'{self.marker_path}').write_text('executed')",))
+
+
+def test_load_pyg_data_reads_data_object(tmp_path):
```
shuo-zhou: Can the new functions be covered by the existing tests for `MultiomicsDataset` and `PolypharmacyDataset`? If yes, no need to add new tests.

kale/loaddata/mnistm.py
```diff
@@ -69,11 +69,11 @@ def __init__(
 
         if self.train:
             self.data, self.targets = torch.load(
-                os.path.join(self.root, self.processed_folder, self.training_file), weights_only=False
+                os.path.join(self.root, self.processed_folder, self.training_file), weights_only=True
```
shuo-zhou, line 72: Please test this example: https://github.com/pykale/pykale/tree/main/examples/digits_dann to confirm the change works.

shuo-zhou (changes requested):

kale/loaddata/pyg_serialization.py
cmanjah: Moved to `kale.loaddata.pyg_serialization` (imports, docs and tests updated), since its only callers are the loaddata datasets and it is PyG-specific.

tests/loaddata/test_pyg_serialization.py
cmanjah: The happy path is also exercised by the existing MultiomicsDataset/PolypharmacyDataset tests. I've kept only the security-specific tests here (arbitrary-object rejection + allow-list purity), which those don't cover and which are the #541 regression guard.

kale/loaddata/mnistm.py
cmanjah: Confirmed: the digits_dann tutorial (executed in CI) passes on 3.10/3.11/3.12 with this change. The MNIST-M .pt is a plain (ByteTensor, labels) tuple, which weights_only=True loads fine. Also added tests/loaddata/test_mnistm.py to cover it outside the notebook.

kale/loaddata/pyg_serialization.py
shuo-zhou: Can this be generalised to `load_graph_data`?

kale/loaddata/mnistm.py
shuo-zhou: Thanks for confirming. Would it be better to add a `weight_only` arg in `__init__`?

docs/source/kale.loaddata.rst
```diff
@@ -70,6 +70,14 @@ kale.loaddata.polypharmacy\_datasets module
    :undoc-members:
    :show-inheritance:
 
+kale.loaddata.pyg\_serialization module
```
shuo-zhou: `pyg_serialization module` is not a suitable API name for me. `graph_access` or `load_graph` could be better.

shuo-zhou (changes requested):
API and function naming to be discussed

kale/loaddata/pyg_serialization.py
cmanjah: Done in f61cc0ac — `load_pyg_data` is now `load_graph_data`, and the whole module moved to `kale/loaddata/graph_access.py` (with `get_pyg_safe_globals` -> `get_graph_safe_globals`). The generic name is the public surface; the implementation stays PyG-specific for now, but the API no longer ties callers to that.

docs/source/kale.loaddata.rst
cmanjah: Done in f61cc0ac — renamed to `kale.loaddata.graph_access`, which matches the `*_access` convention already used in loaddata (`image_access`, `tabular_access`, `signal_access`, `video_access`). Docs and importers updated.

kale/loaddata/mnistm.py
cmanjah: For MNIST‑M, I’d suggest keeping weights_only=True fixed. The processed .pt file is always a simple (ByteTensor, labels) tuple, so weights_only=True reliably loads it, and I haven’t seen a real need for callers to set it to False. Allowing that would also reopen the code‑execution risk that #541 was meant to close.

If you’d still prefer exposing the argument for flexibility, I’m happy to add it with a default of True, but my recommendation is to keep it fixed here.

shuo-zhou (approved):

PR #585, opened 2026-07-20
Use the ResultKeys constant for the concatenated-targets result key

cmanjah:
Follow-up to #575 (merged), addressing its review.

### Description
The producers and consumers of the `"all bins concatenated targets separated"` result key repeated the literal string instead of the existing `ResultKeys.ALL_BINS_CONCAT_TARGETS_SEP` constant — the duplication that let the earlier typo drift. All usages now go through the constant. The value is the same string, so results are unchanged (verified on fixtures).

### Status
Ready

### Types of changes

### DIFF
--- a/kale/evaluate/uncertainty_metrics.py
+++ b/kale/evaluate/uncertainty_metrics.py
@@ -1528,11 +1528,11 @@ def evaluate_bounds(
                     for target_idx in range(len(targets)):
                         fold_all_bins_concat_targets_sep_foldwise[target_idx][idx_bin] = (
                             fold_all_bins_concat_targets_sep_foldwise[target_idx][idx_bin]
-                            + return_dict["all bins concatenated targets separated"][target_idx][idx_bin]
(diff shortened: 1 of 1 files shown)


shuo-zhou (changes requested):
This PR looks okay to me. However, the classes `ColumnNames` and `ResultKey` are too specific. It would be better to refactor these two classes to make them more general. I beleive they can be reused for general tabular input data. @xianyuanliu please let us know if you have any comments.

cmanjah:
Thanks, @shuo-zhou. I agree, but I'd rather not fold that refactor into this PR. It's a two-line change that finishes adopting a constant already defined in `main`, and both classes predate that change. Redesigning them here would make a +/- 2 cleanup into a broader uncertainty pipeline refactor.

shuo-zhou (approved):
Merge this PR and create a new PR to refactor `ColumnNames` and `ResultKey` to fix issue #593

PR #587, opened 2026-07-21
Unify optimizer kwarg validation across trainers

### LINKED ISSUE #544: `[Refactor]` Unify optimizer configuration / kwarg validation across trainers
`configure_optimizers` behaves differently in the two trainer base classes for the same documented
contract:

- `kale/pipeline/domain_adapter.py:420-450` validates kwargs via `validate_kwargs(...)`.
- `kale/pipeline/base_nn_trainer.py:88-113` passes `optim_params` straight through.

A config valid in one pipeline can raise `TypeError` in the other.

- Modules modified: `kale/pipeline/base_nn_trainer.py`, `kale/pipeline/domain_adapter.py` (shared helper).
- Impact on code structure: consolidate optimizer construction into one validated helper.
- Impact on speed: negligible.
- Breaking change: no (only relaxes/standardizes accepted kwargs).

cmanjah:
Fixes #544.

### Description
`BaseNNTrainer.configure_optimizers` and `CNNTransformerTrainer.configure_optimizers` in `kale/pipeline/base_nn_trainer.py` passed `self._optimizer_params["optim_params"]` straight into the torch optimizer constructors, whereas `kale.pipeline.domain_adapter` already wraps them in `validate_kwargs(...)` (`kale/utils/validate.py`). This unifies the base trainers onto the same validated path so both code paths behave identically.

### Behaviour change (please note)
This is **not** a pure cleanup. `validate_kwargs` filters `optim_params` down to keys accepted by the optimizer constructor:
- **Before:** an `optim_params` key the optimizer does not accept raised `TypeError: __init__() got an unexpected keyword argument ...`.
- **After:** that key is silently dropped, matching the long-standing behaviour in `domain_adapter`.

Valid optimizer configurations are entirely unaffected — same optimizer, same hyperparameters. Only the handling of *invalid* keys changes (hard error → silently ignored). I have ticked **Breaking change** because the error semantics of a public method change, even though no valid usage breaks.

### Verified
- Added regression tests `test_configure_optimizers_filters_invalid_optim_params` for both `TestBaseTrainer` and `TestCNNTransformerTrainer`. Reverting only the source change makes both fail with `TypeError` (the old raw path); with the fix both pass.
- `tests/pipeline/test_base_nn_trainer.py` — 22 passed.
- Pinned pre-commit green to convergence; clean tree.
- Import check OK.

### Status
**Ready**

### Types of changes

### DIFF
--- a/kale/pipeline/base_nn_trainer.py
+++ b/kale/pipeline/base_nn_trainer.py
@@ -29,6 +29,7 @@ import torch.nn as nn
 from sklearn.metrics import accuracy_score
 
 import kale.evaluate.metrics as losses
+from kale.utils.validate import validate_kwargs
 
 
 class BaseNNTrainer(pl.LightningModule):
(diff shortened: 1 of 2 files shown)


shuo-zhou (approved):
Look good to me

PR #595, opened 2026-09-23
Move ColumnNames and ResultKeys to kale.evaluate.constants

### LINKED ISSUE #593: [Refactor] Make ColumnNames and ResultKeys reusable beyond quantile binning
`ColumnNames` and `ResultKeys` in `kale/evaluate/uncertainity_metrics.py` are written around landmark uncertainty quantile binning. @shuo-zhou suggested in #585 that they'd be more useful made general, since the same ideas come up with any tabular input. 
This moves shared constants out of a domain-specific module so other evaluation code can use them without importing from uncertainty metrics.  It will be a breaking change for anything importing `ColumnNames` or `ResultKeys` from `kale.evaluate.uncertainty_metrics`

cmanjah:
Fixes #593.

### Description
`ColumnNames` and `ResultKeys` name DataFrame columns and result-dictionary keys, but lived in `kale/evaluate/uncertainty_metrics.py`. Any module wanting the names had to import pandas, torch and torchmetrics along with them: `from kale.evaluate.uncertainty_metrics import ColumnNames` takes 8.4s, against 0.09s from the new module.

This is deliberately a relocation rather than a redesign. @shuo-zhou's point in #585 was that these are too specific to quantile binning, which I agree with, but today they are used almost entirely inside `uncertainty_metrics` itself . Happy to take it further in this PR if @shuo-zhou and @xianyuanliu have a target shape in mind.

### Types of changes


cmanjah:
@claude, review

